PR #21, opened 2016-09-30
Allow customizable credentials for Gini SDK iOS

limfinity:
Using the settings bundle a user can provide his own Gini SDK credenitals.

Relates to: #MOBILE-630


a-szotyori:
LGTM!

PR #26, opened 2016-12-14
Add some notes about iOS10

limfinity:
## Information

Adds a quick note on how to use the Gini SDK with Xcode 8 and iOS 10 to the docs.

## How to test

See the readme for changes.

## Merge information

Nothing to be considered.

## YouTrack Tickets

MOBILE-589, MOBILE-660

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ We are providing example apps for Swift and Objective-C. These apps demonstrate 
 
 Gini Vision Library can either be installed by using CocoaPods or by manually dragging the required files to your project.
 
+**Note on iOS 10:** No matter which option you choose if you want to support iOS 10 you need to specify the `NSPhotoLibraryUsageDescription` key in your `Info.plist` file. And if you're using the [Gini iOS SDK](https://github.com/gini/gini-sdk-ios) you need to add support for "Keychain Sharing" in your entitlements by adding a `keychain-access-groups` value to your entitlements file.
+
 ### CocoaPods
 
 [CocoaPods](http://cocoapods.org) is a dependency manager for Cocoa projects. You can install it with the following command:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -87,6 +87,8 @@ We are providing example apps for Swift and Objective-C. These apps demonstrate
 
 Gini Vision Library can either be installed by using CocoaPods or by manually dragging the required files to your project.
 
+**Note on iOS 10:** No matter which option you choose if you want to support iOS 10 you need to specify the `NSPhotoLibraryUsageDescription` key in your `Info.plist` file. And if you're using the [Gini iOS SDK](https://github.com/gini/gini-sdk-ios) you need to add support for "Keychain Sharing" in your entitlements by adding a `keychain-access-groups` value to your entitlements file.
```
nsobadjiev: Regarding `NSPhotoLibraryUsageDescription `... I remember there were two keys - one for the Camera and one for the Photo Library. If you don't specify them, you either fail review, or you get a crash when accessing that resource.  So, first of all, are you sure we don't actually need `NSCameraUsageDescription`? Secondly, maybe we should also add text explaining why this key is needed

nsobadjiev: Also, we should probably explain why the `keychain-access-groups` key is needed - that the keychain access will fail, resulting in authentication problems.

limfinity: Whoaaa...you are right! I completely messed up the keys. We only need `NSCameraUsageDescription` and `NSPhotoLibraryUsageDescription ` is only for development. Will fix that and add a sentence on why we need it.

limfinity: Maybe we can just point to the integration guide from the Gini SDK for more information.

limfinity:
@nsobadjiev Thanks for the review. Checkout my tiny changes. 🎉

PR #30, opened 2017-01-30
Bugfix mobile 698 camera resizing

nsobadjiev:
# Information

We had the following bug:
* Start the Example project
* Choose the Component API
* Rotate the device (use a real device) to landscape

After completing those steps, the camera preview view wasn't properly aligned.
This pull request fixes the issue by updating several constraints on the video preview and rotating the camera feed as the device rotates.

# How to test

Execute the steps above and verify that the view doesn't cut into the other subviews. **Note** that in landscape the view still respects the 4:3 aspect ratio so it's not going to look really nice. But this is going to be fixed in a subsequent project.

# Merge information

Automatic. Also needs to be cherrypicked in develop and swift-2.3

# Tickets

[MOBILE-698](https://tickets.i.gini.net/issue/MOBILE-698)

PR #31, opened 2017-01-30
Removed the GINI Prefix #MOBILE-699

nsobadjiev:
# Information

Ever since Swift 3 was released, using prefixes for classes is discouraged. Since the Gini Vision Library used "GINI" as prefix, it is a good idea to remove it before version 3.0.X beta becomes official.

For many files and classes, it meant just deleting the prefix, but for some core classes like `GINIVision`, `GINIConfiguration` and `GINIMetadataManager`, the `GINI` was turned into a `Gini` in order to preserve the context information that they are related to the Vision Library and not the whole app.

# How to test

Check if the project follows the new naming convention and functions properly at the same time.

# Merge information

Automatic. However, there might be a conflict with merging #30 "Bugfix mobile 698 camera resizing"

# Tickets

MOBILE-699


GiniVision/Classes/CameraViewController.swift
```diff
@@ -240,15 +240,15 @@ public typealias GINICameraErrorBlock = (_ error: GINICameraError) -> ()
                 var imageData = try inner()
                 
                 // Set meta information in image
-                var manager = GINIMetaInformationManager(imageData: imageData)
+                var manager = GiniMetaInformationManager(imageData: imageData)
```
a-szotyori: I think the Gini prefix can be omitted here.

a-szotyori:
You, sir, may merge. 🖖

PR #34, opened 2017-03-10
Adding UUID and rotation to EXIF

nsobadjiev:
# Introduction

In order to better identified identical (only rotated) images in the backend, we are now adding a UUID to the image's EXIF user comment. Also, the rotation angle in degrees should be included. For example:

`UUID=F26AE2F5-4E12-4CD8-A079-0114A11F3C12,RotDeltaDeg=90`

# How to test

Currently, there are no unit tests, unfortunately. You have to either save the image before uploading and view its EXIF data, or get the uploaded image from the backend

# Merge information

Automatic

### DIFF
--- a/GiniVision/Classes/ImageMetaInformationManager.swift
+++ b/GiniVision/Classes/ImageMetaInformationManager.swift
@@ -273,7 +273,7 @@ internal struct ImageMetaInformationManager {
         var comment = "Platform=\(platform),OSVer=\(osVersion),GiniVisionVer=\(giniVisionVersion),\(userCommentUUID)=\(uuid)"
         if let rotationDegrees = rotationDegrees {
             // normalize the rotation to 0-360
-            let rotation = imageRotation() + rotationDegrees
+            let rotation = imageRotationDeltaDegrees() + rotationDegrees
             let rotationNorm = normalizedDegrees(imageRotation: rotation)
             comment += ",\(userCommentRotation)=\(rotationNorm)"
         }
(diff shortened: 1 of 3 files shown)


GiniVision/Classes/ImageMetaInformationManager.swift
```diff
@@ -255,11 +265,49 @@ internal struct ImageMetaInformationManager {
...
+    }
+    
+    fileprivate func imageUUID() -> String {
+        // if it already has one, reuse it - it shouldn't change
+        // if not, generate it
+        let existingUUID = uuidFromImage()
+        return existingUUID ?? NSUUID().uuidString
+    }
+    
+    fileprivate func imageRotation() -> Int {
```
a-szotyori: I think the function's name should reflect its intent more clearly. Maybe `imageRotationDelta`?

GiniVision/Classes/ImageMetaInformationManager.swift
```diff
@@ -255,11 +265,49 @@ internal struct ImageMetaInformationManager {
...
+        return rotationFromImage() ?? 0
+    }
+    
+    fileprivate func uuidFromImage() -> String? {
+        return self.valueFor(userCommentField: userCommentUUID)
+    }
+    
+    fileprivate func rotationFromImage() -> Int? {
```
a-szotyori: Same as above. `rotationDeltaFromImage()`?

nsobadjiev:
Yes, good point about the method names. I updated them.

PR #37, opened 2017-03-17
Not being able to set custom titles on buttons

nsobadjiev:
# Introduction

This pull request tackles two problems:
* On customisable buttons, it is impossible to override the default image with a custom text
* On customisable navigation buttons, it is impossible to revert to the default iOS back button (an arrow to the left and a title "Back" or the title of the previous screen).

The problem was that client could override images and titles by providing resources with the same name in their project's asset catalog and `Localizable.strings`, respectively. However, if they supplied a custom text, it would still be overridden by setting the default image, since images take precedence to titles in UIKit. 

# How to test

Pick a button from the documentation that is subject to customisation (for instance the back button in the review screen). Add an entry to the `Localizable.strings` or your client or example project with the key used for that button. Run the app and verify that the new text is shown and not the default image.

To test the second part of the fix, set the key to the empty string. Run the app and verify that the default iOS back button is shown (in the review screen example, it you just say "Back", or "Zurück" with german localisation).

# Merge information

Automatic

### DIFF
--- a/GiniVision/Classes/PreferredResource.swift
+++ b/GiniVision/Classes/PreferredResource.swift
@@ -84,3 +84,25 @@ struct PreferredResource {
         return (textSource == .custom && imageSource != .custom)
     }
 }
+
+// MARK Navigation items
+internal extension UIViewController {
+    
+    // Setup the leftNavigationItem property of a UIViewController. It will create an UIBarButtonItem
+    // for it and add it to the navigation bar.
+    // Note that if the preferred title of the resource is the empty string, the leftNavigationItem
+    // will not be set so the default back button is shown
+    func setupLeftNavigationItem(usingResources preferredResources:PreferredResource, selector:Selector) {
+        let buttonText = preferredResources.preferredText
+        if buttonText != nil && !buttonText!.isEmpty {
+            let navButton = GiniBarButtonItem(
+                image: preferredResources.preferredImage,
+                title: buttonText,
+                style: .plain,
+                target: self,
+                action: selector
+            )
+            self.navigationItem.setLeftBarButton(navButton, animated: false)
+        }
+    }
+}
--- a/GiniVision/Classes/PreferredResource.swift
+++ b/GiniVision/Classes/PreferredResource.swift
@@ -1,0 +1,86 @@
+//
+//  PreferredResource.swift
+//  Pods
+//
+//  Created by Nikola Sobadjiev on 16/03/2017.
+//
+//
+
+import UIKit
+
+enum ResourceSource {
+    case unknown
+    case library
+    case custom
+}
+
+/*
+ * PreferredResource typically controls what artwork (image and text) would be used in a UI element
+ * such as a button or a bar button item. Resources in the Gini Vision Library are usually brandable
+ * via UIAppearance, Asset catalogs and localizable strings. For instance, many buttons have a default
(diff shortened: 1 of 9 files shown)


nsobadjiev:
@a-szotyori , I thought for a while what the best way to resolve the issue was. I wanted a solution that is simple and doesn't bring a lot of overheat for the future. But I'm not sure this was the best choice. Any feedback and suggestions regarding the new API is welcome :). If you don't understand anything, just ask me.

GiniVision/Classes/PreferredResource.swift
```diff
@@ -0,0 +1,108 @@
...
+struct PreferredResource {
+    
+    let imageName:String?
+    let text:String?
+    let textComment:String?
+    let appBundle = Bundle.main
+    let libBundle = Bundle(for: GiniVision.self)
+    var imageSource:ResourceSource {
+        if let name = imageName {
+            if UIImage(named: name, in: appBundle, compatibleWith: nil) != nil {
```
nsobadjiev: I couldn't make it happen unfortunately. I thought I could check if there is such file in one bundle or the other. But since we use Asset catalogs, that doesn't work. Maybe I can research a bit more, but so far it seems there's no way to check for files inside an Asset catalog. I think it's because they compile into a simple ".car".

I'm not terribly concerned about the performance of the code thought. After an UIImage is loaded from the file system, it's automatically cached, so any subsequent request for it will have constant complexity. Alternatively, if it's not there, I assume there's hardly any file operation, so that would be really fast as well.

GiniVision/Classes/PreferredResource.swift
```diff
@@ -0,0 +1,108 @@
...
+// MARK Navigation items
+internal extension UIViewController {
+    
+    // Setup the leftNavigationItem property of a UIViewController. It will create an UIBarButtonItem
+    // for it and add it to the navigation bar.
+    // Note that if the preferred title of the resource is the empty string, the leftNavigationItem
+    // will not be set so the default back button is shown
+    func setupLeftNavigationItem(usingResources preferredResources:PreferredResource, selector:Selector) {
+        let buttonText = preferredResources.preferredText
+        if buttonText != nil && !buttonText!.isEmpty {
```
nsobadjiev: Sure, but then we have two nested blocks and 3 levels of indentation. I honestly think it's nore readable without a `guard`.

GiniVision/Classes/ReviewContainerViewController.swift
```diff
@@ -51,26 +47,20 @@ internal class ReviewContainerViewController: UIViewController, ContainerViewCon
         view.backgroundColor = GiniConfiguration.sharedConfiguration.backgroundColor
         
         // Configure back button
-        backButton = GiniBarButtonItem(
-            image: backButtonImage,
-            title: GiniConfiguration.sharedConfiguration.navigationBarReviewTitleBackButton,
```
nsobadjiev: It's a good point. I'll look into keeping the API. We shouldn't break old implementations.

nsobadjiev:
@a-szotyori I fixed the code review issues. I found a way to not change the API for users at all. At this point I think we wont even need an upgrade to version 3.1. Since it's just a bugfix, I think 3.0.X would suffice.

a-szotyori:
Awesome! _I approve this message_ ™ 👍

PR #38, opened 2017-03-21
Example app saves images

a-szotyori:
Sometimes it is useful to get the images that are taken and manipulated by the library (for checking EXIF tags for ex.).

Created a `FileUtil` class which saves `Data` types to a unique file suffixed by a provided string and with the `jpeg` extension. Timestamp is used for uniqueness. It also logs the file path. File name ex.:  `1490114874.42508_for_review.jpeg`

Both Screen API and Component API examples save the original image shown in the Review Screen with the `for_review` suffix and the image shown in the Analysis Screen with the `for_analysis` suffix.

### How to test
Run the example app and take pictures and analyse them with Screen API and Component API. Watch the log to see that the images are saved.

Using iTunes download the images from the device.

### Merge
Automatic


a-szotyori:
I see that `GiniConfiguration` has a property for debug called `debugModeOn` and also a shorthand called `DEBUG` which is used to have a fake camera if it is `true` and there was a `CameraError`. `debugModeOn` is documented as saving images to the camera roll, but it seems we didn`t got to implement that 😁.

`FileUtil` could use the `GiniConfiguration.DEBUG` and do nothing if it is `false`.

Or if you like you could implement saving the images to the camera roll to fulfill the documentation. 😁

PR #39, opened 2017-04-27
Bugfix image compression

nsobadjiev:
# Information

We saw that there was not much change in the uploaded image's quality and size even after changing the JPEG compression level. We debugged that to the `ImageMetainformationManager`. When it was generating compressed images and injecting the EXIF data, there were two conversions between `UIImage` and `NSData`. After the first one, the compression level was lost. This resulted in images with the wrong compression level and it was also wasting processing time.

The fix was to switch to Core Image and compress and annotate the image with EXIF data all in one pass.

# How to test

Start debugging the example app and look at the console logs while uploading the images. The total size in KB should be printed just before the upload starts. To try different compression levels, change the `JPEGDefaultCompression` variable in `ImageMetainformationManager `

# Merge information

Automatic

### DIFF
--- a/Example/GiniVision/ScreenAPIViewController.swift
+++ b/Example/GiniVision/ScreenAPIViewController.swift
@@ -99,6 +99,7 @@ class ScreenAPIViewController: UIViewController {
         
         cancelAnalsyis()
         imageData = data
+        print("Analysing document with size \(Double(data.count) / 1024.0)")
         AnalysisManager.sharedManager.analyzeDocument(withImageData: data, cancelationToken: CancelationToken(), completion: { inner in
             do {
                 guard let response = try inner?(),
--- a/GiniVision/Classes/CameraViewController.swift
+++ b/GiniVision/Classes/CameraViewController.swift
@@ -241,7 +241,7 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
             if GiniConfiguration.DEBUG {
                 // Retrieve image from default image view to make sure image was set and therefor the correct states were checked before.
                 if let image = self.defaultImageView?.image,
-                   let imageData = UIImageJPEGRepresentation(image, 1) {
+                   let imageData = UIImageJPEGRepresentation(image, 0.2) {
                     self.successBlock?(imageData)
                 }
             }
@@ -252,9 +252,9 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
                 var imageData = try inner()
                 
                 // Set meta information in image
-                var manager = ImageMetaInformationManager(imageData: imageData)
+                let manager = ImageMetaInformationManager(imageData: imageData)
                 manager.filterMetaInformation()
-                if let richImageData = manager.imageData(withCompression: 0.2) {
+                if let richImageData = manager.imageData() {
                     imageData = richImageData
                 }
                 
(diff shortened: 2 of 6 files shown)


GiniVision/Classes/ImageMetaInformationManager.swift
```diff
@@ -226,20 +227,22 @@ internal struct ImageMetaInformationManager {
...
-            CGImageDestinationAddImageFromSource(destination, source, i, information as CFDictionary)
+        
+        let targetData = NSMutableData()
+        
+        let destination = CGImageDestinationCreateWithData(targetData, kUTTypeJPEG, 1, nil)!
+        let source = CGImageSourceCreateWithData(image as CFData, nil)
+        information.setValue(compression, forKey: kCGImageDestinationLossyCompressionQuality as String)
+        
+        let count = CGImageSourceGetCount(source!)
+        for _ in 0...count - 1 {
```
a-szotyori: I am pretty sure this iteration is not needed as the destination is created with an image count of 1. We ignore the loop counter and if the source has multiple images we add the first one multiple times to the destination. Don't know what happens in this case as destination was created with an image count of 1...

I think the source would contain in our case only one image* so the loop is unnecessary.

\* **BUT** If we are not sure we should do it as before: get the nr of images from the source, pass the nr to the destination and then add all the images from the source to the destination.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -241,7 +241,7 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
             if GiniConfiguration.DEBUG {
                 // Retrieve image from default image view to make sure image was set and therefor the correct states were checked before.
                 if let image = self.defaultImageView?.image,
-                   let imageData = UIImageJPEGRepresentation(image, 1) {
+                   let imageData = UIImageJPEGRepresentation(image, 0.2) {
```
nsobadjiev: The camera framework doesn't compress the image, yes. However, if you look several lines down, where the real camera image is taken, it immediately passes through `ImageMetainfomationManager` which in turn compresses it. So leaving the `CameraViewController`, the image is always compressed. I changed it to 0.2 because it was a bit strange to see 1.5 MB images while debugging on the simulator.

GiniVision/Classes/ImageMetaInformationManager.swift
nsobadjiev: Yeah, I also thought it was useless, but finally left it just before committing, because it was there in the past. But since you also think it's useless, I'll just remove it to simplicity.

Example/GiniVision/ScreenAPIViewController.swift
```diff
@@ -99,6 +99,7 @@ class ScreenAPIViewController: UIViewController {
         
         cancelAnalsyis()
         imageData = data
+        print("Analysing document with size \(Double(data.count) / 1024.0)")
```
nsobadjiev: Unfortunately, this is how we log in the whole library. Which is an oversight - we need to come up with a proper logging solution. In the meantime, it would probably be annoying for customers to always see a log message about file sizes. I'll surround that with an if.

nsobadjiev: Oh, I just realised that this log message is not in the library, but in the example app. So some additional logging shouldn't be a problem.

Issue #42, opened 2017-08-17
Fix cameraOverlay asset on Landscape


kikettas:
Fix cameraOverlay asset when the device is on landscape

Issue #41, opened 2017-08-16
Adapt UI for tablets


kikettas:
Add or update views, layouts and assets to for the larger screens.

PR #43, opened 2017-08-17
Adapt UI for iPad

kikettas:
Extended the library to support tablets.

Enable rotation on iPad, added iPad constraints on CameraViewController, set default previewView orientation on iPad, and added cameraOverlay image rotation.

### How to test
Execute the tests on a tablet and on a phone.

### Merging
Automatic.

### Ticket
#41, #42, #44, #47


kikettas:
I have not finished, I found a bug on the OnboardingViewController

Example/GiniVision.xcodeproj/project.pbxproj
```diff
@@ -360,6 +360,7 @@
 						CreatedOnToolsVersion = 6.3.1;
 						DevelopmentTeam = JA825X8F7Z;
 						LastSwiftMigration = 0810;
+						ProvisioningStyle = Automatic;
```
nsobadjiev: What are all these changes in the project file?

Example/Podfile.lock
```diff
@@ -7,7 +7,7 @@ PODS:
   - Bolts/Tasks (1.1.5)
   - Gini-iOS-SDK (0.3.2):
     - Bolts (~> 1.1.0)
-  - GiniVision (3.0.4)
+  - GiniVision (3.0.5)
```
nsobadjiev: Why did you change the version now?

GiniVision/Classes/CameraViewController.swift
```diff
@@ -68,11 +69,19 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
+                                           UIDeviceOrientation.landscapeRight: AVCaptureVideoOrientation.landscapeLeft,
+                                           UIDeviceOrientation.landscapeLeft: AVCaptureVideoOrientation.landscapeRight,
+                                           UIDeviceOrientation.portraitUpsideDown: AVCaptureVideoOrientation.portraitUpsideDown]
+    
     
     // Properties
     fileprivate var camera: Camera?
     fileprivate var cameraState = CameraState.notValid
     
+
```
nsobadjiev: Why a second empty line? We don't usually do this.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -84,7 +93,10 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
         return UIImageNamedPreferred(named: "cameraCaptureButtonActive")
     }
     fileprivate var cameraOverlayImage: UIImage? {
-        return UIImageNamedPreferred(named: "cameraOverlay")
+        guard let image = UIImageNamedPreferred(named: "cameraOverlay"), let cgImage = image.cgImage else{
```
nsobadjiev: Isn't the cameraOverlay image too small to scale it up to iPad sizes? Doesn't it get blurry?

Ideally, I'd like to remove it and add the rectangle programmatically. We can copy the code from GiniSwitch.

The problem is that the image is also customisable for clients so we need a major version update to fix it. @a-szotyori , do you think  we can do it now or in the near future?

GiniVision/Classes/CameraViewController.swift
```diff
@@ -131,13 +143,15 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
         if let validCamera = camera {
             cameraState = .valid
             previewView.session = validCamera.session
-            (previewView.layer as! AVCaptureVideoPreviewLayer).videoGravity = AVLayerVideoGravityResizeAspect
+            (previewView.layer as! AVCaptureVideoPreviewLayer).videoGravity = AVLayerVideoGravityResizeAspectFill
```
nsobadjiev: Are you sure about this `AVLayerVideoGravityResizeAspectFill`?

GiniVision/Classes/CameraViewController.swift
```diff
@@ -194,14 +209,12 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
-        // in UIDeviceOrientation and AVCaptureVideoOrientation. Refer to their documentaion for more info
-        let orientationsMapping = [UIDeviceOrientation.portrait: AVCaptureVideoOrientation.portrait,
-                                   UIDeviceOrientation.landscapeRight: AVCaptureVideoOrientation.landscapeLeft,
-                                   UIDeviceOrientation.landscapeLeft: AVCaptureVideoOrientation.landscapeRight,
-                                   UIDeviceOrientation.portraitUpsideDown: AVCaptureVideoOrientation.portraitUpsideDown]
+        
         let orientation = UIDevice.current.orientation
         (previewView.layer as? AVCaptureVideoPreviewLayer)?.connection?.videoOrientation = orientationsMapping[orientation]!
+        
+        // Set cameraOverlay image once device has rotated
```
nsobadjiev: This comment doesn't give me much information. I know you're setting the image, but it doesn't explain why.

Either remove it, or write an explanation.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -194,14 +209,12 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
+        cameraOverlay.image = cameraOverlayImage
```
nsobadjiev: This means that every time you rotate, a new image is generated. It's not ideal.

Also, it's kind of a side effect. You are setting the image, which would normally make no difference, but you are using the side effect of the getter always generating a new image and considering the device orientation.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -241,7 +254,7 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
             if GiniConfiguration.DEBUG {
                 // Retrieve image from default image view to make sure image was set and therefor the correct states were checked before.
                 if let image = self.defaultImageView?.image,
-                   let imageData = UIImageJPEGRepresentation(image, 0.2) {
+                    let imageData = UIImageJPEGRepresentation(image, 0.2) {
```
nsobadjiev: This space here displeases me.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -303,60 +316,86 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
-        let superview = self.view
-        
-        // Preview view
+        addPreviewViewConstraints()
+        addCameraOverlayConstraints()
+        addControlsViewConstraints()
+        addCaptureButtonConstraints()
+    }
+    
+    fileprivate func addPreviewViewConstraints(){
```
nsobadjiev: Usually we put a space between the ")" and "{". I'd prefer it if you did the same to keep it consistent.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -303,60 +316,86 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
-        
-        // required constraints - make sure the preview doesn't expand into other views or off-screen
-        ConstraintUtils.addActiveConstraint(item: superview!, attribute: .bottom, relatedBy: .greaterThanOrEqual, toItem: previewView, attribute: .bottom, multiplier: 1, constant: 0, priority: 1000)
-        ConstraintUtils.addActiveConstraint(item: previewView, attribute: .left, relatedBy: .greaterThanOrEqual, toItem: superview, attribute: .left, multiplier: 1, constant: 0, priority: 1000)
-        ConstraintUtils.addActiveConstraint(item: superview!, attribute: .right, relatedBy: .greaterThanOrEqual, toItem: previewView, attribute: .right, multiplier: 1, constant: 0, priority: 1000)
-        
-        ConstraintUtils.addActiveConstraint(item: previewView, attribute: .width, relatedBy: .equal, toItem: previewView, attribute: .height, multiplier: 3/4, constant: 0)
-        ConstraintUtils.addActiveConstraint(item: previewView, attribute: .centerX, relatedBy: .equal, toItem: superview, attribute: .centerX, multiplier: 1, constant: 0)
+        if UIDevice.current.userInterfaceIdiom == .pad{
```
nsobadjiev: This check for tablet or phone is a bit ugly. I'm not sure we can do it in any other way, but there might be. Can you try to think of an alternative?

GiniVision/Classes/CameraViewController.swift
```diff
@@ -303,60 +316,86 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
+            ConstraintUtils.addActiveConstraint(item: previewView, attribute: .top, relatedBy: .equal, toItem: self.view, attribute: .top, multiplier: 1, constant: 0)
+            ConstraintUtils.addActiveConstraint(item: previewView, attribute: .bottom, relatedBy: .equal, toItem: self.view, attribute: .bottom, multiplier: 1, constant: 0)
+            ConstraintUtils.addActiveConstraint(item: previewView, attribute: .leading, relatedBy: .equal, toItem: self.view, attribute: .leading, multiplier: 1, constant: 0)
+            ConstraintUtils.addActiveConstraint(item: previewView, attribute: .trailing, relatedBy: .equal, toItem: controlsView, attribute: .leading, multiplier: 1, constant: 0)
+        }else{
```
nsobadjiev: Again, we tend to put spaces between "else" statements.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -303,60 +316,86 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
         cameraOverlay.translatesAutoresizingMaskIntoConstraints = false
+        
         // All constraints here have a priority less than required to make sure they don't get broken
         // when the view gets too small
-        ConstraintUtils.addActiveConstraint(item: cameraOverlay, attribute: .top, relatedBy: .equal, toItem: previewView, attribute: .top, multiplier: 1, constant: 0, priority: 999)
+        ConstraintUtils.addActiveConstraint(item: cameraOverlay, attribute: .top, relatedBy: .equal, toItem: previewView, attribute: .top, multiplier: 1, constant: 23, priority: 999)
         ConstraintUtils.addActiveConstraint(item: cameraOverlay, attribute: .trailing, relatedBy: .equal, toItem: previewView, attribute: .trailing, multiplier: 1, constant: -23, priority: 999)
-        ConstraintUtils.addActiveConstraint(item: cameraOverlay, attribute: .bottom, relatedBy: .equal, toItem: previewView, attribute: .bottom, multiplier: 1, constant: 0, priority: 999)
+        ConstraintUtils.addActiveConstraint(item: cameraOverlay, attribute: .bottom, relatedBy: .equal, toItem: previewView, attribute: .bottom, multiplier: 1, constant: -23, priority: 999)
```
nsobadjiev: Why is this change from 0 to 23?

Example/Dev-Config.xcconfig
```diff
@@ -1,2 +1,2 @@
 #include "Dev-Keys.xcconfig"
-GCC_PREPROCESSOR_DEFINITIONS = $(inherited) GINI_CLIENT_ID=@\"$(GINI_CLIENT_ID)\" GINI_CLIENT_SECRET=@\"$(GINI_CLIENT_SECRET)\"
\ No newline at end of file
+GCC_PREPROCESSOR_DEFINITIONS = $(inherited) GINI_CLIENT_ID=@\"$(GINI_CLIENT_ID)\" GINI_CLIENT_SECRET=@\"$(GINI_CLIENT_SECRET)\"
```
nsobadjiev: It might be my personal preferences, but before I commit, I always check all the changes and remove changes like that. They make it distracting when you're trying to review the code later because you don't only see the important changes but also loads of "noise".

GiniVision/Classes/CameraViewController.swift
```diff
@@ -50,12 +50,13 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
  - note: Component API only.
  */
+
 @objc public final class CameraViewController: UIViewController {
     
     /**
-    Image view used to display a camera overlay like corners or a frame.
-    Use public methods `showCameraOverlay` and `hideCameraOverlay` to control visibility of overlay.
-    */
+     Image view used to display a camera overlay like corners or a frame.
```
nsobadjiev: What are all these extra spaces everywhere in this file?

Example/GiniVision.xcodeproj/project.pbxproj
kikettas: I think it has been generated automatically when I used Gini provisioning profile. Does it need to be manual rather than automatic for non-release purposes?

Example/Podfile.lock
kikettas: I did not change it, I just run `pod install` so maybe it took the last version

GiniVision/Classes/CameraViewController.swift
a-szotyori: I can create higher resolution images for the iPad from the photoshop assets file (as I did for Android). 

I prefer continuing using images to be consistent. Clients would only need to provide new iPad images, which can be expected when updating to support iPads.

GiniVision/Classes/CameraViewController.swift
kikettas: In the code there isn't that space, weird...

GiniVision/Classes/CameraViewController.swift
kikettas: Yes, I'm sorry... fixed

GiniVision/Classes/CameraViewController.swift
```diff
@@ -131,13 +143,15 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
         if let validCamera = camera {
             cameraState = .valid
             previewView.session = validCamera.session
-            (previewView.layer as! AVCaptureVideoPreviewLayer).videoGravity = AVLayerVideoGravityResizeAspect
+            (previewView.layer as! AVCaptureVideoPreviewLayer).videoGravity = AVLayerVideoGravityResizeAspectFill
+            (previewView.layer as! AVCaptureVideoPreviewLayer).connection?.videoOrientation = UIDevice.current.userInterfaceIdiom == .pad ?
+                orientationsMapping[UIDevice.current.orientation] ?? .portrait :
+                .portrait
```
a-szotyori: I prefer clarity over brevity, because imho the time you save writing this someone will spend in the future and with interest.

I recommend using a method to set the video orientation where the logic is made easy to read with if conditional.

I like the elvis operator, but how can it happen that the orientationsMapping does not have a corresponding item for the device orientation?

GiniVision/Classes/CameraViewController.swift
kikettas: I think so, because otherwise the previewView shows black spaces on the sides when it is on portrait mode, and it does not look very good. The image taken is good and the extractions are the same.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -303,60 +316,86 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
-        ConstraintUtils.addActiveConstraint(item: controlsView, attribute: .top, relatedBy: .equal, toItem: previewView, attribute: .bottom, multiplier: 1, constant: 0, priority: 750)
-        ConstraintUtils.addActiveConstraint(item: controlsView, attribute: .trailing, relatedBy: .equal, toItem: superview, attribute: .trailing, multiplier: 1, constant: 0)
-        ConstraintUtils.addActiveConstraint(item: controlsView, attribute: .bottom, relatedBy: .equal, toItem: superview, attribute: .bottom, multiplier: 1, constant: 0)
-        ConstraintUtils.addActiveConstraint(item: controlsView, attribute: .leading, relatedBy: .equal, toItem: superview, attribute: .leading, multiplier: 1, constant: 0)
-        ConstraintUtils.addActiveConstraint(item: controlsView, attribute: .height, relatedBy: .greaterThanOrEqual, toItem: captureButton, attribute: .height, multiplier: 1.1, constant: 0)
-        
-        // Capture button
+    }
+    
+    fileprivate func addCaptureButtonConstraints(){
```
a-szotyori: Space: the final frontier ... or what to put before curly brackets ... 🖖

GiniVision/Classes/CameraViewController.swift
a-szotyori: I agree with Nikola, the intent is not clear. You could use a method like `cameraOverlayImageForOrientation(orientation)` (adapt the name to swift conventions ;) )

Also to avoid image creation you should cache the images for portrait and landscape.

GiniVision/Classes/CameraViewController.swift
kikettas: Displeases me too, but It is Xcode indentation not a white space.

GiniVision/Classes/GiniNavigationViewController.swift
```diff
@@ -42,7 +42,7 @@ internal extension GiniNavigationViewController {
     }
     
     override var supportedInterfaceOrientations : UIInterfaceOrientationMask {
-        return UIInterfaceOrientationMask.portrait
+        return UIDevice.current.userInterfaceIdiom == .pad ? .all : .portrait
```
a-szotyori: We have an `Extensions.swift` file, you should add `UIDevice.current.userInterfaceIdiom == .pad` (or the solution for detecting iPads) into a helper method (`isiPad()`?) there.

GiniVision/Classes/CameraViewController.swift
kikettas: It was not ugly for me, but I came up with a shorter solution.

GiniVision/Classes/CameraViewController.swift
kikettas: Because it had not any space with both the top and the bottom and it did not look very consistent.

GiniVision/Classes/GiniNavigationViewController.swift
kikettas: That was actually what came up to my mind and added an hour ago :)

kikettas:
I finished with the changes you requested. I fixed also an issue with NotAuthorizedView (appears when there is no camera).

@a-szotyori @nsobadjiev

GiniVision/Classes/Camera.swift
```diff
@@ -66,10 +66,11 @@ internal class Camera {
                 return completion({ _ in throw CameraError.noInputDevice })
             }
             // Set the orientation accoding to the current orientation of the device
+            
             if let orientation = AVCaptureVideoOrientation(self.motionManager.currentOrientation) {
                 connection.videoOrientation = orientation
             } else {
-                connection.videoOrientation = .portrait
+                connection.videoOrientation = AVCaptureVideoOrientation(UIApplication.shared.statusBarOrientation)
```
nsobadjiev: having a dependency on `UIApplication` is not ideal, IMHO. Also, this is the status bar orientation, not the device orientation. I don't remember correctly, but I think there might be a case when this is not 100% what you want.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -71,11 +71,16 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
     fileprivate var defaultImageView: UIImageView?
     // note that in the mapping left and right are reversed. That's because landscape values have different meanings
     // in UIDeviceOrientation and AVCaptureVideoOrientation. Refer to their documentaion for more info
-    fileprivate let orientationsMapping = [UIDeviceOrientation.portrait: AVCaptureVideoOrientation.portrait,
+    fileprivate let deviceOrientationsMapping = [UIDeviceOrientation.portrait: AVCaptureVideoOrientation.portrait,
                                            UIDeviceOrientation.landscapeRight: AVCaptureVideoOrientation.landscapeLeft,
                                            UIDeviceOrientation.landscapeLeft: AVCaptureVideoOrientation.landscapeRight,
                                            UIDeviceOrientation.portraitUpsideDown: AVCaptureVideoOrientation.portraitUpsideDown]
     
+    fileprivate let interfaceOrientationsMapping = [UIInterfaceOrientation.portrait: AVCaptureVideoOrientation.portrait,
```
nsobadjiev: Why are BOTH of these two arrays necessary? 
Was it because there was a mismatch between landscape left and right on iPad?

Since this is becoming bigger and bigger, maybe it makes sense to extract it somewhere?

GiniVision/Classes/CameraViewController.swift
```diff
@@ -95,7 +100,12 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
         guard let image = cameraOverlayImage, let cgImage = image.cgImage else {
             return nil
         }
-        return UIImage(cgImage: cgImage , scale: 1.0, orientation: UIDevice.current.orientation.isLandscape ? .right : UIImageOrientation.up)
+        if UIDevice.current.orientation.isFlat {
+            return UIImage(cgImage: cgImage , scale: 1.0, orientation: UIApplication.shared.statusBarOrientation.isLandscape ? .right : UIImageOrientation.up)
+        } else{
```
nsobadjiev: always with these spaces ;(

GiniVision/Classes/CameraViewController.swift
```diff
@@ -95,7 +100,12 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
+            return UIImage(cgImage: cgImage , scale: 1.0, orientation: UIDevice.current.orientation.isLandscape ? .right : UIImageOrientation.up)
```
nsobadjiev: Duplicates the return statement above. Only the orientation is different.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -95,7 +100,12 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
         guard let image = cameraOverlayImage, let cgImage = image.cgImage else {
             return nil
         }
-        return UIImage(cgImage: cgImage , scale: 1.0, orientation: UIDevice.current.orientation.isLandscape ? .right : UIImageOrientation.up)
+        if UIDevice.current.orientation.isFlat {
+            return UIImage(cgImage: cgImage , scale: 1.0, orientation: UIApplication.shared.statusBarOrientation.isLandscape ? .right : UIImageOrientation.up)
```
nsobadjiev: Why does one return statement use the status bar orientation and the other, the device orientation?

GiniVision/Classes/CameraViewController.swift
```diff
@@ -95,7 +100,12 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
         guard let image = cameraOverlayImage, let cgImage = image.cgImage else {
             return nil
         }
-        return UIImage(cgImage: cgImage , scale: 1.0, orientation: UIDevice.current.orientation.isLandscape ? .right : UIImageOrientation.up)
+        if UIDevice.current.orientation.isFlat {
```
nsobadjiev: Why does it matter if the device is face up or face down?

GiniVision/Classes/CameraViewController.swift
a-szotyori: If the statusBarOrientation delivers correct results for the UI all the time, we should use only that.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -210,7 +220,7 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
         super.viewWillTransition(to: size, with: coordinator)
         
         let orientation = UIDevice.current.orientation
-        (previewView.layer as? AVCaptureVideoPreviewLayer)?.connection?.videoOrientation = orientationsMapping[orientation]!
+        (previewView.layer as? AVCaptureVideoPreviewLayer)?.connection?.videoOrientation = deviceOrientationsMapping[orientation] ?? interfaceOrientationsMapping[UIApplication.shared.statusBarOrientation]!
```
nsobadjiev: I really don't understand these orientation things. Also, another dependency on UIApplication.

a-szotyori: Here, too, we should only use the statusBarOrientation (and interfaceOrientationsMapping) if there is no case where it delivers erronous results.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -283,7 +293,11 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
     
     fileprivate func getVideoOrientation(forDeviceOrientation orientation:UIDeviceOrientation) -> AVCaptureVideoOrientation{
         if UIDevice.current.isIpad {
-            return orientationsMapping[UIDevice.current.orientation] ?? .portrait
+            if UIDevice.current.orientation.isFlat { // This only could happen on initialization
```
nsobadjiev: This comment is difficult to understand for me. Either explain it a little better or remove it.

The same thing with the `isFlat` and the mappings repeats. Maybe you should extract it somewhere.

GiniVision/Classes/CameraViewController.swift
a-szotyori: Also shouldn't we use the getVideoOrientaiton() method here?

GiniVision/Classes/CameraViewController.swift
```diff
@@ -283,7 +293,11 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
+                return interfaceOrientationsMapping[UIApplication.shared.statusBarOrientation] ?? .portrait
```
a-szotyori: Same here as before we should use only the statusBarOrientation, if there are no side-effects.

GiniVision/Classes/OnboardingContainerViewController.swift
```diff
@@ -90,6 +90,15 @@ internal class OnboardingContainerViewController: UIViewController, ContainerVie
         displayContent(contentController)
     }
     
+    public override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+        // Current onboarding page needs to be centered during transition (after ScrollView changes its frame)
+        coordinator.animate(alongsideTransition: { (_) in
+            (self.contentController as? OnboardingViewController)?.centerTo(page: self.pageControl.currentPage)
+        })
+        
+        super.viewWillTransition(to: size, with: coordinator)
```
nsobadjiev: Are you sure the super call needs to be at the end?

GiniVision/Classes/OnboardingContainerViewController.swift
```diff
@@ -90,6 +90,15 @@ internal class OnboardingContainerViewController: UIViewController, ContainerVie
         displayContent(contentController)
     }
     
+    public override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+        // Current onboarding page needs to be centered during transition (after ScrollView changes its frame)
+        coordinator.animate(alongsideTransition: { (_) in
+            (self.contentController as? OnboardingViewController)?.centerTo(page: self.pageControl.currentPage)
```
a-szotyori: Would it make sense to use a weak self reference? Just asking.

GiniVision/Classes/OnboardingPage.swift
```diff
@@ -84,9 +84,9 @@ import UIKit
         // Image view
         imageView.translatesAutoresizingMaskIntoConstraints = false
         ConstraintUtils.addActiveConstraint(item: imageView, attribute: .top, relatedBy: .equal, toItem: contentView, attribute: .top, multiplier: 1, constant: 0)
-        ConstraintUtils.addActiveConstraint(item: imageView, attribute: .width, relatedBy: .lessThanOrEqual, toItem: nil, attribute: .width, multiplier: 1, constant: 204)
+        ConstraintUtils.addActiveConstraint(item: imageView, attribute: .width, relatedBy: .lessThanOrEqual, toItem: nil, attribute: .width, multiplier: 1, constant: 612)
```
nsobadjiev: 612... that' quite unusual but very specific. How did you calculate it? And if it's calculated, why not calculate it dynamically?

GiniVision/Classes/OnboardingPage.swift
```diff
@@ -84,9 +84,9 @@ import UIKit
...
         ConstraintUtils.addActiveConstraint(item: imageView, attribute: .width, relatedBy: .greaterThanOrEqual, toItem: nil, attribute: .width, multiplier: 1, constant: 75)
-        ConstraintUtils.addActiveConstraint(item: imageView, attribute: .height, relatedBy: .lessThanOrEqual, toItem: nil, attribute: .height, multiplier: 1, constant: 120)
+        ConstraintUtils.addActiveConstraint(item: imageView, attribute: .height, relatedBy: .lessThanOrEqual, toItem: nil, attribute: .height, multiplier: 1, constant: 360)
```
nsobadjiev: Same here

GiniVision/Classes/OnboardingViewController.swift
```diff
@@ -123,6 +123,17 @@ import UIKit
...
+    
+    /**
+     Center page in case it is not centered (i.e after rotation)
+     
+     */
+    public func centerTo(page:Int) {
+        var offset = scrollView.contentOffset
+        offset.x = scrollView.frame.width * CGFloat(page)
+                
+        scrollView.setContentOffset(offset, animated: true)
```
nsobadjiev: Doesn't the page change size too?

GiniVision/Classes/CameraViewController.swift
kikettas: As we discussed before, statusBarOrientation always returns a value even if status bar is hidden. The deviceOrientationMapping is needed because UIInterfaceOrientation has not changed when viewWillTransition is called. And the interfaceOrientationMapping is needed because It could be the case that there is no UIDeviceOrientation (and you need it on initialization).

GiniVision/Classes/CameraViewController.swift
kikettas: Yes, fixed it.

GiniVision/Classes/OnboardingContainerViewController.swift
kikettas: I think that they animation won't outlive the viewController, but in any case I will check it.

GiniVision/Classes/Camera.swift
kikettas: I found out that when the device is on flat position (neither landscape or portrait) there is no UIDeviceOrientation, and in that case it would not be correct to fallback always to portrait, given that it is not correct and we have another orientation reference (UIInterfaceOrientation).

kikettas: What do you suggest to get the UIInterfaceOrientation without having the UIApplication dependency.

GiniVision/Classes/CameraViewController.swift
kikettas: Aa I told to Alpar, the deviceOrientationMapping is needed because UIInterfaceOrientation has not changed when viewWillTransition is called. And the interfaceOrientationMapping is needed because It could be the case that there is no UIDeviceOrientation (and you need it on initialization).

GiniVision/Classes/CameraViewController.swift
kikettas: On initialization there is not UIDeviceOrientation, so you need a reference to your current Orientation in order to set the correct orientation to cameraOverlay.

GiniVision/Classes/CameraViewController.swift
kikettas: I explained it above.

GiniVision/Classes/CameraViewController.swift
kikettas: I explained it above.

GiniVision/Classes/OnboardingContainerViewController.swift
kikettas: I put it before because the animation could start if super method is called before. I just checked that there is no problem, so I put the super called before in order to be consistent.

GiniVision/Classes/OnboardingPage.swift
kikettas: It is three times the old value (the old value is taken from the max image width size). I have tested with twice the old value on iPad and I think it looks better being three times the old value.
Following the same criteria from the iPhone, it should be 810 instead of 612 but I tested it and IMO It looks too big.

Anyway, the problem it is not with the max value (612) but with the mini which has not been modified.

GiniVision/Classes/OnboardingPage.swift
kikettas: Explained above.

GiniVision/Classes/OnboardingViewController.swift
kikettas: Yes it does. It is filling the container before the rotation.

kikettas:
I found out that there is a problem with orientation settings on Info.plist and UIDevice orientation. Apparently although you set only portrait mode the Info.plist, _UIDevice.current.orientation_ still returns current physical orientation. 
The solution has been to remove all references to UIDevice.current.orientation (including deviceOrientationMapping) and to rewrite _viewWillTransition_ method, executing the inner code alongside with the animation.

Example/GiniVision.xcodeproj/project.pbxproj
nsobadjiev: No, automatic is fine, but I don't like seeing unrelated changes in a commit. This PR is about Tablet support and changing the provisioning settings is unrelated.

Example/Podfile.lock
nsobadjiev: So, then 3.0.5 should be the right one. However, again, it's an unrelated change.

GiniVision/Classes/CameraViewController.swift
nsobadjiev: Yeah, ok, we will have to refactor those constraints and ideally not have so many hardcoded constants, but for now it's good.

GiniVision/Classes/OnboardingPage.swift
nsobadjiev: I'm not against the value itself. I'm just wondering why we have to hardcode values like that. My intuition is that a strange value like 204 or 612 is maybe calculated in some way and might be also derived dynamically in code. I'll look into it a little more.

GiniVision/Classes/ImageMetaInformationManager.swift
```diff
@@ -287,6 +288,13 @@ internal class ImageMetaInformationManager {
         return existingUUID ?? NSUUID().uuidString
     }
     
+    fileprivate func deviceOrientation() -> String {
+        if UIApplication.shared.statusBarOrientation.isPortrait {
```
nsobadjiev: Another dependency on `UIApplication`...

GiniVision/Classes/ImageMetaInformationManager.swift
```diff
@@ -270,7 +270,8 @@ internal class ImageMetaInformationManager {
         let osVersion = UIDevice.current.systemVersion
         let giniVisionVersion = GiniVision.versionString
         let uuid = imageUUID()
-        var comment = "Platform=\(platform),OSVer=\(osVersion),GiniVisionVer=\(giniVisionVersion),\(userCommentContentId)=\(uuid)"
+        let devOrientation = deviceOrientation()
```
nsobadjiev: I don't understand the purpose of this variable. You have a property AND a function to calculate it. I assume it is because you want to initialise the comment variable implicitly on the next line. However, what if the orientation changes in the meantime?

GiniVision/Classes/ImageMetaInformationManager.swift
```diff
@@ -287,6 +288,13 @@ internal class ImageMetaInformationManager {
         return existingUUID ?? NSUUID().uuidString
     }
     
+    fileprivate func deviceOrientation() -> String {
```
nsobadjiev: Normally, I'd say that a method about a device orientation shouldn't be in a class that deals with EXIF tags. However, I know that `ImageMetaDataManager` desperately needs to be re-written.

So no action required here, just wanted to point it out.

GiniVision/Classes/OnboardingPage.swift
kikettas: I didn't like to harcode the values neither. At first I created some static properties for them, do you think It would be better or it would be enough just adding a comment above explaining?

GiniVision/Classes/ImageMetaInformationManager.swift
kikettas: I was just trying to be consistent to previous code, because it may be more clear for others to read. Yes, I thought about that possibility but It is quite unlikely to happen, isn't it?

GiniVision/Classes/ImageMetaInformationManager.swift
kikettas: As I commented above, what do you suggest to solve this? I was thinking about some ideas, but none of them convinced me.

GiniVision/Classes/ImageMetaInformationManager.swift
a-szotyori: Talking about this with Enrique we realized, that getting the current device orientation every time the user comment is updated can lead to false values: take pic in portrait, review it and rotate it in landscape -> device orientation will be changed to landscape, although the pic was taken in portrait. Thank you @nsobadjiev! 👍👍👍👍

GiniVision/Classes/Extensions.swift
```diff
@@ -151,3 +152,125 @@ internal extension UIDevice {
...
+
+internal extension Collection where Iterator.Element == CFString {
+    
+    var strings: [ String ] {
+        return self.map { $0 as String }
+    }
+    
+}
+
+internal extension NSMutableDictionary {
```
a-szotyori: This is MetaInformationManager implementation specific and should remain in that file.

GiniVision/Classes/ImageMetaInformationManager.swift
```diff
@@ -160,16 +39,36 @@ internal class ImageMetaInformationManager {
         kCGImagePropertyTIFFSoftware
     ]
     
-    var imageData: Data?
-    var metaInformation: MetaInformation?
-    
     // user comment fields
     let userCommentRotation = "RotDeltaDeg"
     let userCommentContentId = "ContentId"
+    let userCommentPlatform = "Platform"
```
a-szotyori: 👍 Nice that you extracted the remaining constants, too!

GiniVision/Classes/Extensions.swift
kikettas: Moved again as we discussed!

nsobadjiev:
Awesome! I'll merge after Travis succeeds.

https://media2.giphy.com/media/EzUnXqOTGmVaw/giphy.gif

Issue #53, opened 2017-09-01
Fix travis 'pod install' error


kikettas:
Travis fails sometimes when there are some jobs running in parallel, returning this error:

`pod install --project-directory= --repo-update failed`

PR #54, opened 2017-09-01
Fix travis 'pod install' error

kikettas:
Sometimes when travis tries to run jobs in parallel, it fails returning `pod install --project-directory=Example --repo-update failed` error. 
Apparently it is due to`xcode8.2` image is not working well with cocoapods dependencies. With `xcode9` image it seems to be solved. 

### How to test
Restart travis build

### Merging
Automatic.

### Ticket
#53

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode8.2
+osx_image: xcode9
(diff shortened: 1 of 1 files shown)


kikettas:
I have the same opinion @nsobadjiev, but I think Xcode 9 is quite stable now.

Issue #51, opened 2017-08-31
File import with 'Open with...'


kikettas:
Provide a way to handle files imported with 'Open with...' feature.

PR #56, opened 2017-09-05
File import with "Open with..."

kikettas:
From external apps there is always an option to export a resource to another different app, so this feature enables the Example app to be discoverable by other apps and adds ways to handle incoming files by GVL. 
For the Example app is only possible to handle files from ScreenAPIViewController, since it is a GiniVisionDelegate.

### How to test
Try export an image or pdf from other app to Example app.

### Merging
Automatic.

### Ticket 
Issue #51

### DIFF
--- a/Example/GiniVision/AppDelegate.swift
+++ b/Example/GiniVision/AppDelegate.swift
@@ -39,7 +39,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
     
     func application(_ application: UIApplication, open url: URL, sourceApplication: String?, annotation: Any) -> Bool {
-        // If the current root view controller is not the ScreenAPIViewController (a GiniVisionDelegate), it won't do anything.
+        // This is only a shortcut for demo purposes since if the current root view controller is not
+        // the ScreenAPIViewController (a GiniVisionDelegate), it won't do anything.
         guard let navVC = window?.rootViewController as? UINavigationController, let screenAPIVC = navVC.viewControllers.first as? ScreenAPIViewController else {
             return false
         }
@@ -68,3 +69,4 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
     
 }
+
--- a/Example/GiniVision/AppDelegate.swift
+++ b/Example/GiniVision/AppDelegate.swift
@@ -39,7 +39,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
     
     func application(_ application: UIApplication, open url: URL, sourceApplication: String?, annotation: Any) -> Bool {
-        // If the current root view controller is not the ScreenAPIViewController, it won't do anything.
+        // If the current root view controller is not the ScreenAPIViewController (a GiniVisionDelegate), it won't do anything.
         guard let navVC = window?.rootViewController as? UINavigationController, let screenAPIVC = navVC.viewControllers.first as? ScreenAPIViewController else {
             return false
         }
--- a/Example/GiniVision/AppDelegate.swift
+++ b/Example/GiniVision/AppDelegate.swift
@@ -45,18 +45,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             return false
         }
         
-        // 1. Create GiniConfiguration
-        let giniConfiguration = GiniConfiguration()
-        giniConfiguration.debugModeOn = true
-        giniConfiguration.navigationBarItemTintColor = UIColor.white
-        
-        // 2. Read data imported from url
+        // 1. Read data imported from url
         let data = try? Data(contentsOf: url)
         
-        // 3. Create the Gini Vision Library view controller, set a delegate object and pass in the configuration object
-        let vc = GiniVision.viewController(withDelegate: screenAPIVC, withConfiguration: giniConfiguration, importedFile: data)
+        // 2. Create the Gini Vision Library view controller, set a delegate object and pass in the configuration object
+        let vc = screenAPIVC.giniScreenAPI(withImportedFile: data)
         
-        // 4. Present the Gini Vision Library Screen API modally
+        // 3. Present the Gini Vision Library Screen API modally
         screenAPIVC.present(vc, animated: true, completion: nil)
         
         return true
--- a/Example/GiniVision/AppDelegate.swift
+++ b/Example/GiniVision/AppDelegate.swift
@@ -38,6 +38,28 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         return true
     }
     
+    func application(_ application: UIApplication, open url: URL, sourceApplication: String?, annotation: Any) -> Bool {
+        // If the current root view controller is not the ScreenAPIViewController not action done
+        guard let navVC = window?.rootViewController as? UINavigationController, let screenAPIVC = navVC.viewControllers.first as? ScreenAPIViewController else {
+            return false
+        }
+        
+        let giniConfiguration = GiniConfiguration()
+        giniConfiguration.debugModeOn = true
+        giniConfiguration.navigationBarItemTintColor = UIColor.white
+        
+        _ = url.startAccessingSecurityScopedResource()
+        let data = try? Data(contentsOf: url)
+        url.stopAccessingSecurityScopedResource()
+        // 2. Create the Gini Vision Library view controller, set a delegate object and pass in the configuration object
+        let vc = GiniVision.viewController(withDelegate: screenAPIVC, withConfiguration: giniConfiguration, importedFile: data)
+        
+        // 3. Present the Gini Vision Library Screen API modally
+        screenAPIVC.present(vc, animated: true, completion: nil)
+        
+        return true
+    }
+    
     func populateSettingsPage() {
         UserDefaults.standard.setValue(GiniVision.versionString, forKey: kSettingsGiniVisionVersionKey)
         UserDefaults.standard.setValue(Bundle.main.object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String, forKey: kSettingsExampleAppVersionKey)
--- a/Example/GiniVision/AppDelegate.swift
+++ b/Example/GiniVision/AppDelegate.swift
@@ -39,22 +39,23 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
     
     func application(_ application: UIApplication, open url: URL, sourceApplication: String?, annotation: Any) -> Bool {
-        // If the current root view controller is not the ScreenAPIViewController not action done
+        // If the current root view controller is not the ScreenAPIViewController, it won't do anything.
         guard let navVC = window?.rootViewController as? UINavigationController, let screenAPIVC = navVC.viewControllers.first as? ScreenAPIViewController else {
             return false
         }
         
+        // 1. Create GiniConfiguration
         let giniConfiguration = GiniConfiguration()
         giniConfiguration.debugModeOn = true
         giniConfiguration.navigationBarItemTintColor = UIColor.white
         
-        _ = url.startAccessingSecurityScopedResource()
+        // 2. Read data imported from url
         let data = try? Data(contentsOf: url)
-        url.stopAccessingSecurityScopedResource()
-        // 2. Create the Gini Vision Library view controller, set a delegate object and pass in the configuration object
+        
+        // 3. Create the Gini Vision Library view controller, set a delegate object and pass in the configuration object
         let vc = GiniVision.viewController(withDelegate: screenAPIVC, withConfiguration: giniConfiguration, importedFile: data)
         
-        // 3. Present the Gini Vision Library Screen API modally
+        // 4. Present the Gini Vision Library Screen API modally
         screenAPIVC.present(vc, animated: true, completion: nil)
         
         return true
--- a/Example/GiniVision/AppDelegate.swift
+++ b/Example/GiniVision/AppDelegate.swift
@@ -48,11 +48,22 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // 1. Read data imported from url
         let data = try? Data(contentsOf: url)
         
-        // 2. Create the Gini Vision Library view controller, set a delegate object and pass in the configuration object
-        let vc = screenAPIVC.giniScreenAPI(withImportedFile: data)
+        // 2. Create alert which allows user to open imported file either with ScreenAPI or ComponentAPI
+
+        let alertViewController = UIAlertController(title: "Importierte Datei", message: "Möchten Sie die importierte Datei mit dem ScreenAPI oder ComponentAPI verwenden?", preferredStyle: .alert)
         
-        // 3. Present the Gini Vision Library Screen API modally
-        screenAPIVC.present(vc, animated: true, completion: nil)
+        alertViewController.addAction(UIAlertAction(title: "Screen API", style: .default) { _ in
+            screenAPIVC.present(screenAPIVC.giniScreenAPI(withImportedFile: data), animated: true, completion: nil)
+        })
+        
+        alertViewController.addAction(UIAlertAction(title: "Component API", style: .default) { _ in
+            if let componentAPI = screenAPIVC.giniComponentAPI(withImportedFile: data) {
+                screenAPIVC.present(componentAPI, animated: true, completion: nil)
+            }
+        })
+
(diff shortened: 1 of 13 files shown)


GiniVision/Classes/GiniVision.swift
```diff
@@ -87,12 +87,20 @@ import UIKit
...
      - parameter delegate: An instance conforming to the `GiniVisionDelegate` protocol.
+     - parameter importedFile:  A document which comes from other source different than CameraViewController (optional)
      
      - returns: A presentable navigation view controller.
      */
-    public class func viewController(withDelegate delegate: GiniVisionDelegate) -> UIViewController {
-        let cameraContainerViewController = CameraContainerViewController()
-        let navigationController = GiniNavigationViewController(rootViewController: cameraContainerViewController)
+    public class func viewController(withDelegate delegate: GiniVisionDelegate, importedFile:Data? = nil) -> UIViewController {
+        let viewController:UIViewController = {
```
a-szotyori: I like this approach as a programmer. As a maintainer or contributor I would avoid it. The parantheses are easy to miss and you have to go back and check again after you see them and understand that this is a block call for conditional instantiation. Also imho a block is a bit of an overkill for this case. I don't like being a party 💩-er, but in this case I would vote for a simple approach:
```
let viewController:UIViewController;
if let importedFile = importedFile {
    viewController = ReviewContainerViewController(imageData: importedFile)
} else {
    viewController = CameraContainerViewController()
}
```

GiniVision/Classes/ReviewContainerViewController.swift
```diff
@@ -20,7 +20,9 @@ internal class ReviewContainerViewController: UIViewController, ContainerViewCon
...
     
-    fileprivate let backButtonResources = PreferredButtonResource(image: "navigationReviewBack", title: "ginivision.navigationbar.review.back", comment: "Button title in the navigation bar for the back button on the review screen", configEntry: GiniConfiguration.sharedConfiguration.navigationBarReviewTitleBackButton)
+    fileprivate lazy var backButtonResources = PreferredButtonResource(image: "navigationReviewBack", title: "ginivision.navigationbar.review.back", comment: "Button title in the navigation bar for the back button on the review screen", configEntry: GiniConfiguration.sharedConfiguration.navigationBarReviewTitleBackButton)
+    
+    fileprivate lazy var closeButtonResources = PreferredButtonResource(image: "navigationCameraClose", title: "ginivision.navigationbar.camera.close", comment: "Button title in the navigation bar for the close button on the camera screen", configEntry: GiniConfiguration.sharedConfiguration.navigationBarCameraTitleCloseButton)
```
a-szotyori: The strings for `title:` and `comment:` should be updated as we are in the Review Screen here. We should add a new config entry for this button's title: `navigationBarReviewTitleCloseButton`

GiniVision/Classes/ReviewContainerViewController.swift
```diff
@@ -78,10 +77,35 @@ internal class ReviewContainerViewController: UIViewController, ContainerViewCon
         displayContent(contentController)
     }
     
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        
+        // Configure back button. Needs to be done here because otherwise the navigation controller would be nil
+        guard let _ = navigationItem.leftBarButtonItem else {
```
a-szotyori: I'm not sure about using `guard` here. As far as I can tell we are not enforcing any preconditions here, there are no statements after the `guard`. We check the existence of the bar button item and if it's missing we add it.

I propose doing it the "old school" way, which conveys the intent on first sight:

```
if navigationItem.leftBarButtonItem == nil {
    setupLeftNavigationItem(usingResources: backButtonPreferredResource(), selector:#selector(back))
}
```

GiniVision/Classes/ReviewContainerViewController.swift
```diff
@@ -78,10 +77,35 @@ internal class ReviewContainerViewController: UIViewController, ContainerViewCon
...
+            return
+        }
+    }
+    
     @IBAction func back() {
         let delegate = (navigationController as? GiniNavigationViewController)?.giniDelegate
         delegate?.didCancelReview?()
-        _ = navigationController?.popViewController(animated: true)
+        if let navVC = navigationController {
+            if navVC.viewControllers.count > 1 {
```
a-szotyori: This check for > 1 view controllers in the stack is used in two places and should be extracted into a method to DRY it up and also make the code more readable.

Example/GiniVision/Info.plist
```diff
@@ -6,6 +6,36 @@
...
 	<string>GiniVision</string>
+    <key>CFBundleDocumentTypes</key>
+    <array>
+        <dict>
+            <key>CFBundleTypeIconFiles</key>
+            <array/>
+            <key>CFBundleTypeName</key>
+            <string>PDF</string>
+            <key>CFBundleTypeRole</key>
+            <string>Editor</string>
```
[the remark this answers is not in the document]
kikettas: Yes, I was thinking about not adding it but I found some answers on SO that said that it was needed. But yes, it should be `Viewer`.

GiniVision/Classes/GiniVision.swift
kikettas: There is no discussion. I don't know way I chose the difficult approach, it is much simpler what you suggest.
No `;` needed btw ☕ 😉

GiniVision/Classes/ReviewContainerViewController.swift
kikettas: I thought about it but I wasn't sure that it was worth for only one line. But since one of the methods has an `else`,  I cannot extract it into a `isFirstViewControllerInTheStack` method/computedProperty.

So, do you think that It is worth to move `navVC.viewControllers.count > 1` to a method?

GiniVision/Classes/ReviewContainerViewController.swift
kikettas: Yes you are right. Too much Swift in my head... 😅

GiniVision/Classes/ReviewContainerViewController.swift
a-szotyori: What about this? I added some guards to make it more obvious that a navigation controller is required.
```
override func viewWillAppear(_ animated: Bool) {
        super.viewWillAppear(animated)
        
        setupLeftBarButton()
    }
    
    fileprivate func setupLeftBarButton() {
        guard let navController = navigationController else {
            return
        }
        
        // Configure back button. Needs to be done here because otherwise the navigation controller would be nil
        if navigationItem.leftBarButtonItem == nil {
            setupLeftNavigationItem(usingResources: backButtonPreferredResource(forNavController: navController), selector:#selector(back))
        }
    }
    
    @IBAction func back() {
        let delegate = (navigationController as? GiniNavigationViewController)?.giniDelegate
        delegate?.didCancelReview?()
        
        handleBackPressed()
    }
    
    fileprivate func handleBackPressed() {
        guard let navController = navigationController else {
            return
        }
    
        if firstViewController(inNavController: navController) {
            navController.dismiss(animated: true, completion: nil)
        } else {
            _ = navController.popViewController(animated: true)
        }
    }
    
    fileprivate func backButtonPreferredResource(forNavController navController:UINavigationController) -> PreferredButtonResource {
        if firstViewController(inNavController: navController) {
            return closeButtonResources
        }
        return backButtonResources
    }
    
    fileprivate func firstViewController(inNavController navController:UINavigationController ) -> Bool {
        if navController.viewControllers.count == 1 {
            return true
        }
        return false
    }
```

Example/GiniVision/AppDelegate.swift
```diff
@@ -38,6 +38,29 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         return true
     }
     
+    func application(_ application: UIApplication, open url: URL, sourceApplication: String?, annotation: Any) -> Bool {
+        // If the current root view controller is not the ScreenAPIViewController, it won't do anything.
```
nsobadjiev: Why does it have to be the `ScreenAPIViewController `?

Example/GiniVision/AppDelegate.swift
```diff
@@ -38,6 +38,29 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        guard let navVC = window?.rootViewController as? UINavigationController, let screenAPIVC = navVC.viewControllers.first as? ScreenAPIViewController else {
+            return false
+        }
+        
+        // 1. Create GiniConfiguration
+        let giniConfiguration = GiniConfiguration()
+        giniConfiguration.debugModeOn = true
+        giniConfiguration.navigationBarItemTintColor = UIColor.white
```
nsobadjiev: I don't really like this. The Gini Configuration is typically created elsewhere so this code is duplicated.

And setting the `navigationBarItemTintColor` is kind of strange. Why would the Open With feature care about the tint color. Set it somewhere where it would apply for all cases.

GiniVision/Classes/GiniVision.swift
```diff
@@ -87,12 +87,19 @@ import UIKit
      - note: Screen API only.
      
      - parameter delegate: An instance conforming to the `GiniVisionDelegate` protocol.
+     - parameter importedFile:  A document which comes from other source different than CameraViewController (optional)
```
nsobadjiev: A document which comes from ***a** source different than...

GiniVision/Classes/GiniVision.swift
```diff
@@ -104,12 +111,13 @@ import UIKit
      
      - parameter delegate:      An instance conforming to the `GiniVisionDelegate` protocol.
      - parameter configuration: The configuration to set.
+     - parameter importedFile:  A document which comes from other source different than CameraViewController (optional)
```
nsobadjiev: see above.

GiniVision/Classes/ReviewContainerViewController.swift
```diff
@@ -78,10 +77,54 @@ internal class ReviewContainerViewController: UIViewController, ContainerViewCon
         displayContent(contentController)
     }
     
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        
+        setupLeftBarButton()
```
nsobadjiev: If you put this in `viewDidLoad`, I guess you would be able to skip the `navigationItem.leftBarButtonItem == nil` in the method.

Is there a reason you put it in `viewWillAppear`?

GiniVision/Classes/ReviewContainerViewController.swift
```diff
@@ -78,10 +77,54 @@ internal class ReviewContainerViewController: UIViewController, ContainerViewCon
...
+    }
+    
+    fileprivate func backButtonPreferredResource(forNavController navController:UINavigationController) -> PreferredButtonResource {
+        if self.firstViewController(inNavController: navController) {
+            return closeButtonResources
+        }
+        return backButtonResources
+    }
+    
+    fileprivate func firstViewController(inNavController navController:UINavigationController ) -> Bool {
```
nsobadjiev: A name like `firstViewController ` makes me think it'd return a view controller. Isn't `isFirstViewController` better and more swifty?

GiniVision/Classes/ReviewContainerViewController.swift
```diff
@@ -78,10 +77,54 @@ internal class ReviewContainerViewController: UIViewController, ContainerViewCon
...
+        if navController.viewControllers.count == 1 {
```
nsobadjiev: Wouldn't comparing with `navController.rootViewController` be better?

Example/GiniVision/AppDelegate.swift
kikettas: Because it is the GiniVisionDelegate, maybe It should be explained.

Example/GiniVision/AppDelegate.swift
kikettas: I just copied the code from the ScreenAPIViewController in order to have the same results. It is not even necessary to add any special attribute to GiniConfiguration, but It was just for having the same results.

GiniVision/Classes/ReviewContainerViewController.swift
kikettas: Yes, `navigationController` is still `nil` in `viewDidLoad`.

GiniVision/Classes/ReviewContainerViewController.swift
kikettas: `UINavigationController` doesn't have a `rootViewController` property. The idea is to check if there is only one viewController in order to change the back icon and handle back action.

Example/GiniVision/AppDelegate.swift
```diff
@@ -39,7 +39,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
     
     func application(_ application: UIApplication, open url: URL, sourceApplication: String?, annotation: Any) -> Bool {
-        // If the current root view controller is not the ScreenAPIViewController, it won't do anything.
+        // If the current root view controller is not the ScreenAPIViewController (a GiniVisionDelegate), it won't do anything.
```
nsobadjiev: That doesn't make it more understandable at all. 
If you want, just say that this is a shortcut for demo purposes.

Example/GiniVision/AppDelegate.swift
nsobadjiev: `I just copied the code` Exactly the problem. I rest my case.

kikettas: I don't understand. What is the problem here? Should I avoid duplicated code in this case?

nsobadjiev: Yeah, find a way to not have this here. Apply the configuration once in a place where it would apply for all cases (Open With or normal app start).

a-szotyori: You could move this into the ScreenAPIViewController to a public method similar to `easyLaunchGiniVision` named maybe `easyLaunchGiniVision(withImportedFile:)` or something. It's better to have Screen API stuff in one class.

Also for the Component API you could add a similar method to `ComponentAPIAnalysisViewController` and allow clients to opt for trying out the Component API here. A simple enum property with a switch would look nice I think, just an idea: 
```
switch giniVisionAPI {
case .Screen:
    print("Screen")
case .Component:
    print("Component")
}
```

kikettas: I came up with the idea of showing an alert which allows to choose between ScreenAPI and ComponentAPI when a file is imported. I don't know if the title and message of the alert are correct (I translated it to german using GT).

Example/GiniVision/ComponentAPIReviewViewController.swift
```diff
@@ -63,6 +63,10 @@ class ComponentAPIReviewViewController: UIViewController {
         // Cancel analysis process to avoid unnecessary network calls.
         if parent == nil {
             AnalysisManager.sharedManager.cancelAnalysis()
+        } else {
+            if let nav = parent as? UINavigationController, let cameraContainer = nav.viewControllers[nav.viewControllers.count - 2] as? ComponentAPICameraViewController {
```
nsobadjiev: That statement makes me really nervous. Especially the `nav.viewControllers[nav.viewControllers.count - 2]`.
Since this is in the sample app, which uses a storyboard, can't you use a rewind segue to pop to a specific view controller? I can help you with it, if you want.

kikettas: I tried to use the `unwind` segue so I could be more consistent with the segues, but I had to use a custom button for the UINavigationController (without back arrow) and add an IBAction in the viewController (when there was already a `didMoveTo...` method). 
What do you think?

nsobadjiev: Even with a custom button you should be able to use unwind. Also, you can add the unwind segue as an outlet and perform it programmatically.

kikettas: Yes, I meant that maybe It could be overkill when we already have the `didMoveTo...` method, and we would also lose the _back arrow_ on back button.

kikettas:
Although this pull request hasn't been approved by al reviewers, It will be merge and the requested changes on Example will be tackle on subsequent PR.

Closes #51

kikettas:
Although this pull request hasn't been approved by al reviewers, It will be merge and the requested changes on Example will be tackle on subsequent PR.

Closes #51

Issue #50, opened 2017-08-30
File import from other apps 


kikettas:
Import files from other apps and from camera roll

Issue #52, opened 2017-08-31
File type check


kikettas:
Apart from file extension, it is also necessary to check the file type inside the file itself.

PR #57, opened 2017-09-05
File import and file type check

kikettas:
Since not only an image can be analyzed but also a PDF, there was too many changes. It's been created a new `GiniVisionDocument` protocol and two classes which implement it, `GiniImageDocument` and `GiniPDFDocument`.
Given that the code was very coupled to `imageData` references, it has been necessary to change several parameters on initializers and methods, adapting them to the new classes mentioned above.

### How to test
Run the app and import a pdf or image from the Camera screen. If you run an iOS 11 device, try dropping a PDF or image (JPEG, PNG, GIF or TIFF) from other app like _Files_.

### Merging
Automatic.

### Ticket 
Issue #50 and #52


kikettas:
@a-szotyori @nsobadjiev 

You can observe that there are some comments where error handling is needed. I have preferred to not continue with it until you review all these changes since that depends on it.

GiniVision/Classes/ReviewViewController.swift
```diff
@@ -136,6 +156,24 @@ public typealias ReviewErrorBlock = (_ error: ReviewError) -> ()
...
+     
+     - returns: A view controller instance allowing the user to review a picture of a document.
+     */
+    
+    @nonobjc
+    @available(*, deprecated)
+    public convenience init(_ imageData:Data, success: @escaping ReviewSuccessBlock, failure: @escaping ReviewErrorBlock) {
+        self.init(GiniImageDocument(data: imageData), successBlock: { (document,_) in
+            success(document.data)
+        }, failureBlock: failure as! GVCameraScreenFailureBlock)
```
a-szotyori: You should wrap it instead of casting it.

GiniVision/Classes/ReviewViewController.swift
```diff
@@ -162,6 +200,22 @@ public typealias ReviewErrorBlock = (_ error: ReviewError) -> ()
...
+            do {
+                try currentDocument.validate()
+            } catch let error as DocumentValidationError {
+                failureBlock!(error)
+            } catch _ {
+                failureBlock!(DocumentValidationError.unknown)
+            }
+            
+            // If the document is a PDF, go to Analysis screen
+            if currentDocument.type == .PDF {
```
a-szotyori: This decision should be made outside of the `ReviewViewController`, either when launching the library (open with) or when a file was picked in the `CameraViewController`.

Example/GiniVision/Info.plist
```diff
@@ -26,6 +26,8 @@
 	<true/>
 	<key>NSCameraUsageDescription</key>
 	<string>Um Dokumente mit deinem Gerät zu scannen, erlaube den Zugriff auf deine Kamera.  📸 -&gt; 📄 -&gt; 🎉</string>
+	<key>NSPhotoLibraryUsageDescription</key>
+	<string>Um Dokumente mit deinem Gerät zu scannen, erlaube den Zugriff auf deine KameraRoll.  🏞 -&gt; 📄 -&gt; 🎉</string>
```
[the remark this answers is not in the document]
kikettas: I knew that Google Translate would not work... thank you!  👍

Example/GiniVision_Example.entitlements
```diff
@@ -2,6 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>com.apple.developer.icloud-container-identifiers</key>
```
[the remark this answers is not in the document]
kikettas: I wrote all things we need to mention on the docs.

GiniVision/Classes/GiniVisionDocument.swift
```diff
@@ -0,0 +1,62 @@
...
+    var data:Data { get }
+    var previewImage:UIImage? { get }
+    
+    init(data:Data)
+    func checkType() throws
+}
+
+// MARK: GiniDocumentType
+
+@objc public enum GiniDocumentType:Int {
```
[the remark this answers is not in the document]
kikettas: When I realized that `GiniDocument`was used on `GiniSDK`, I changed it to `GiniVisionDocument`but I forgot to do it for the `enum`.

Thanks!

GiniVision/Classes/GiniVisionDocument.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    // MARK: File size check
+    
+    fileprivate func maxFileSizeExceeded(forData data:Data) -> Bool {
+        let bcf = ByteCountFormatter()
+        bcf.allowedUnits = [.useMB] // optional: restricts the units to MB only
+        bcf.countStyle = .file
+        bcf.includesUnit = false
+        
+        if let fileSize = NumberFormatter().number(from: bcf.string(fromByteCount: Int64(data.count)))?.doubleValue {
```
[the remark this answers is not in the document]
kikettas: There is no doubt that It could be simplified. 👍

GiniVision/Classes/GiniVision.swift
```diff
@@ -17,19 +17,19 @@ import UIKit
...
     
     /**
-     Called when the user has taken an image.
+     Called when the user has taken an image or picked a PDF.
      
-     - parameter imageData: JPEG image data including meta information.
+     - parameter fileData: JPEG image data including meta information or PDF data
      */
-    func didCapture(_ imageData: Data)
+    func didCapture(_ fileData: Data)
```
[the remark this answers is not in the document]
kikettas: Do you mean adding a `isImported` property on the `successBlock` so It could be switched on CameraContainerViewController?

GiniVision/Classes/GiniVision.swift
```diff
@@ -17,19 +17,19 @@ import UIKit
...
     
     /**
      Called when the user has reviewed the image and potentially rotated it to the correct orientation.
      
-     - parameter imageData: JPEG image data including eventually updated meta information.
+     - parameter fileData:  JPEG image data including eventually updated meta information or PDF Data
      - parameter changes:   Indicates whether `imageData` was altered.
     */
-    func didReview(_ imageData: Data, withChanges changes: Bool)
+    func didReview(_ fileData: Data, withChanges changes: Bool)
```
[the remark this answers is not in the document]
kikettas: Yes, I thought about this. But, where would you validate de document? Because when the file is imported with the "Open with..." option it does not go through CameraScreen.

GiniVision/Classes/ReviewViewController.swift
```diff
@@ -16,13 +16,27 @@ import UIKit
 public typealias ReviewSuccessBlock = (_ imageData: Data) -> ()
 
 /**
+ Block which will be executed each time the user rotates a picture. It contains the JPEG representation of the image including meta information about the rotated image. In the case of a PDF, it should finish once it has been validated.
+ 
+ - note: Component API only.
+ */
+public typealias GVReviewScreenSuccessBlock = (_ document: GiniVisionDocument, _ shouldFinish:Bool) -> ()
```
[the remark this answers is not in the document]
kikettas: Explained the reason above.

GiniVision/Classes/ReviewViewController.swift
kikettas: Related to previous comments.

GiniVision/Classes/ReviewViewController.swift
kikettas: Yes, actually there is a bug here. Thanks! 👍

GiniVision/Classes/GiniVision.swift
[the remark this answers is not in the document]
a-szotyori: Yes.

GiniVision/Classes/GiniVision.swift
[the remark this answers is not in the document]
a-szotyori: Yeah, that's true. I was thinking about this and realized, there is no reason to call this delegate method for pdfs since this serves to notify clients, that a review finished and if there were changes cancel the analysis and start it with the new Data. Pdfs cannot change so no need to call this. What do you think?

GiniVision/Classes/ReviewViewController.swift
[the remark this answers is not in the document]
a-szotyori: I see. I think the name `shouldFinish` could more clearly convey the intent by renaming it to `shouldProceedToAnalysisScreen`.

GiniVision/Classes/GiniVision.swift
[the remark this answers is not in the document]
kikettas: Yes, you are right. Because it is only call when "Weiter" is pressed  👍

GiniVision/Classes/CameraViewController.swift
```diff
@@ -143,7 +143,7 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> ()
         filePickerManager.didPickFile = { [unowned self] document in
             do {
                 try document.validate()
-                self.successBlock?(document)
+                self.successBlock?(document, true)
```
a-szotyori: Named parameters are so nice for readability (the one thing I really liked about Obj-C). What do you think about using it here? 😍 👉 `self.successBlock(withDocument: document, isImported: true)`

kikettas: Yes, I love them. But unfortunately you cannot use labels on function types 😩

GiniVision/Classes/FilePickerManager.swift
```diff
@@ -12,6 +12,7 @@ import MobileCoreServices
 internal final class FilePickerManager:NSObject {
         
     var didPickFile:((GiniVisionDocument) -> ()) = { _ in }
+    fileprivate var acceptedDocumentTypes = [kUTTypePDF as String] + GiniImageDocument.acceptedImageTypes
```
a-szotyori: Could move `[kUTTypePDF as String]` to `GiniPDFDocument.acceptedType` as it belongs to that classes responsibility.

Example/GiniVision/ComponentAPIAnalysisViewController.swift
```diff
@@ -19,7 +19,7 @@ class ComponentAPIAnalysisViewController: UIViewController {
     /**
      The image data of the captured document to be reviewed.
      */
-    var imageData: Data!
+    var document: GiniVisionDocument!
```
nsobadjiev: I guess the controller cannot work without a document. Would it be possible to not have it as an optional?

Example/GiniVision/ComponentAPIAnalysisViewController.swift
```diff
@@ -92,7 +92,7 @@ class ComponentAPIAnalysisViewController: UIViewController {
         hideErrorButton()
         
         // Retry analysis of the document.
-        AnalysisManager.sharedManager.analyzeDocument(withImageData: imageData, cancelationToken: CancelationToken(), completion: nil)
+        AnalysisManager.sharedManager.analyzeDocument(withData: document.data, cancelationToken: CancelationToken(), completion: nil)
```
nsobadjiev: I guess `sharedManager` is not really Swift-3-like. We might consider renaming it to `shared`. But in another PR.

Example/GiniVision/ComponentAPIReviewViewController.swift
```diff
@@ -22,19 +22,19 @@ class ComponentAPIReviewViewController: UIViewController {
     /**
      The image data of the captured document to be reviewed.
      */
-    var imageData: Data!
+    var document: GiniVisionDocument!
     
-    fileprivate var originalData: Data?
+    fileprivate var originalDocument: GiniVisionDocument?
```
nsobadjiev: And here... the original is an optional but the output is force unwrapped. I'd really like to unify it.

Example/GiniVision_Example.entitlements
```diff
@@ -2,6 +2,10 @@
...
+	<array/>
+	<key>com.apple.developer.ubiquity-kvstore-identifier</key>
```
nsobadjiev: What is this? Keychain?

Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -3,7 +3,7 @@ import XCTest
 
 class CameraViewControllerTests: XCTestCase {
     
-    var vc = CameraViewController(success: { _ in }, failure: { _ in })
+    var vc = CameraViewController(successBlock: { _ in }, failureBlock: { _ in })
```
nsobadjiev: Why did you choose `*Block` instead of just success or failure? Just curious...

Example/Tests/GiniVisionDocumentTests.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+    
+    func testNotExcedeedMaxFileSize() {
+        // Given
+        let lowerThanOrEqualTo10MBData = generateFakeData(megaBytes: 10)
+        
+        // When
+        let fakeDocument = GiniPDFDocument(data: lowerThanOrEqualTo10MBData)
+
+        // Then
```
nsobadjiev: Given... when... then... I really don't think it makes it more readable. It's quite obvious already. Unless you have a strong opinion against it, I'd remove it.

Example/Tests/GiniVisionDocumentTests.swift
```diff
@@ -0,0 +1,62 @@
...
+        XCTAssertThrowsError(try fakeDocument.validate(), "should indicate that is not a valid document") { error in
+            XCTAssert(error as? DocumentValidationError != DocumentValidationError.exceededMaxFileSize, "should indicate that max file size has been exceeded")
+        }
+    }
+    
+    func testImageValidation() {
+        // Given
+        let testBundle = Bundle(for: type(of: self))
+        let image = UIImage(named: "tabBarIconHelp", in: testBundle, compatibleWith: nil)
```
nsobadjiev: I ran into the same thing in the Switch SDK. Getting an image from the framework's bundle is annoying (since you always have to remember to get it from the framework's bundle, not the main one). Can you extract it in a utility method/class so you don't have to repeat yourself in the future?

Example/Tests/GiniVisionDocumentTests.swift
```diff
@@ -0,0 +1,62 @@
...
+
+        // When
+        let imageDocument = GiniImageDocument(data: UIImagePNGRepresentation(image!)!)
+        
+        // Then
+        XCTAssertNoThrow(try imageDocument.validate(), "should not throw an exception given that it is a valid image")
```
nsobadjiev: I personally put more verbose text for my assertions. Something that gives you all the information, so you don't have to look at the context (which test was it, on which class...?). So in this case, it would be something like:

> Valid images should validate without throwing an exception.

Example/Tests/GiniVisionDocumentTests.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+    
+    fileprivate func generateFakeData(megaBytes lengthInMB:Int) -> Data {
```
nsobadjiev: I really like this method. I'm just wondering if it performs well. Does it take a lot of time to generate random data? I wouldn't want to see tests taking a lot of time because if it.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -17,13 +17,28 @@ import AVFoundation
 public typealias CameraSuccessBlock = (_ imageData: Data) -> ()
 
 /**
- Block which will be executed if an error occurs on the camera screen. It contains a camera specific error.
+ Block which will be executed when the camera screen successfully takes a picture or pick a document/picture. It contains the JPEG representation of the image including meta information about the image, or the PDF Data. Should be indicated if the document was imported or not.
+ 
+ - note: Component API only.
+ */
+public typealias CameraScreenSuccessBlock = (_ document: GiniVisionDocument, _ isImported:Bool) -> ()
```
nsobadjiev: Why would clients care if it was imported or not? I'm not sure I agree with that...

GiniVision/Classes/CameraViewController.swift
```diff
@@ -17,13 +17,28 @@ import AVFoundation
 public typealias CameraSuccessBlock = (_ imageData: Data) -> ()
 
 /**
- Block which will be executed if an error occurs on the camera screen. It contains a camera specific error.
+ Block which will be executed when the camera screen successfully takes a picture or pick a document/picture. It contains the JPEG representation of the image including meta information about the image, or the PDF Data. Should be indicated if the document was imported or not.
```
nsobadjiev: `Should be indicated if the document was imported or not`
That's not good wording... sounds funny. But I hope it won't be needed - see my next comment.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -17,13 +17,28 @@ import AVFoundation
...
+
+/**
+ Block which will be executed if an error occurs on the camera. It contains a camera specific error.
  
  - note: Component API only.
  */
 public typealias CameraErrorBlock = (_ error: CameraError) -> ()
 
 /**
+ Block which will be executed if an error occurs on the camera screen.
```
nsobadjiev: Block *that*

GiniVision/Classes/CameraViewController.swift
```diff
@@ -77,6 +93,7 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
     // Properties
     fileprivate var camera: Camera?
     fileprivate var cameraState = CameraState.notValid
+    fileprivate var filePickerManager: FilePickerManager
```
nsobadjiev: I'm not sure I like the idea of having image pickers in a **Camera** ViewController. I would expect the CameraViewController to only have a camera.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -67,6 +82,7 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
     fileprivate var controlsView  = UIView()
     fileprivate var previewView   = CameraPreviewView()
     fileprivate var captureButton = UIButton()
+    fileprivate var documentProviderButton = UIButton()
```
nsobadjiev: Is document provider the button for choosing an image from the camera roll? The name doesn't make it clear.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -103,23 +120,36 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
         super.init(nibName: nil, bundle: nil)
         
         // Set callback
-        successBlock = success
-        errorBlock = failure
+        self.successBlock = successBlock
+        self.failureBlock = failureBlock
+        
+        // Configure file picker
+        filePickerManager.didPickFile = { [unowned self] document in
```
nsobadjiev: `didPickFile` sounds more like a delegate method, not a closure. For closures, I see that typically you use "*Block", so it's inconsistent.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -103,23 +120,36 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
+            do {
+                try document.validate()
```
nsobadjiev: Just curious.. why did you choose to use exceptions rather than returning errors here?

GiniVision/Classes/CameraViewController.swift
```diff
@@ -103,23 +120,36 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
...
      
-     - parameter success: Success block to be executed when image was taken.
+     - parameter success: Success block to be executed when document was picked or image was taken.
      - parameter failure: Error block to be executed if an error occurred.
      
-     - returns: A view controller instance allowing the user to take a picture.
+     - returns: A view controller instance allowing the user to take a picture or pick a document.
      */
-    public init(success: @escaping CameraSuccessBlock, failure: @escaping CameraErrorBlock) {
+    public init(successBlock: @escaping CameraScreenSuccessBlock, failureBlock: @escaping CameraScreenFailureBlock) {
```
nsobadjiev: This is the longest method ever... can you extract some functionality away from it?

GiniVision/Classes/CameraViewController.swift
```diff
@@ -256,7 +314,8 @@ public typealias CameraErrorBlock = (_ error: CameraError) -> ()
                 // Retrieve image from default image view to make sure image was set and therefor the correct states were checked before.
                 if let image = self.defaultImageView?.image,
```
nsobadjiev: I'm not sure we need this default debug image anymore. In the simulator, we can use the camera roll to import an image. But I'm also ok with leaving it here.

GiniVision/Classes/Extensions.swift
```diff
@@ -9,6 +9,7 @@
 import UIKit
 import AVFoundation
 import ImageIO
+import MobileCoreServices
```
nsobadjiev: Why do you need that `MobileCoreServices `?

GiniVision/Classes/FilePickerManager.swift
```diff
@@ -0,0 +1,105 @@
...
+// MARK: UIDocumentPickerDelegate
+
+extension FilePickerManager: UIDocumentPickerDelegate {
+    func documentPicker(_ controller: UIDocumentPickerViewController, didPickDocumentAt url: URL) {
+        do {
+            _ = url.startAccessingSecurityScopedResource()
+            let data = try Data(contentsOf: url)
+            url.stopAccessingSecurityScopedResource()
+            
+            if data.isPDF {
```
nsobadjiev: Can you find a way to avoid these if statements? Like a factory method or something?

GiniVision/Classes/GiniImageDocument.swift
```diff
@@ -0,0 +1,48 @@
...
+//  GiniVision
+//
+//  Created by Enrique del Pozo Gómez on 9/4/17.
+//  Copyright © 2017 Gini GmbH. All rights reserved.
+//
+
+import Foundation
+import MobileCoreServices
+
+final public class GiniImageDocument: NSObject, GiniVisionDocument {
```
nsobadjiev: Why does it have to be a `NSObject`?

GiniVision/Classes/ReviewContainerViewController.swift
```diff
@@ -23,20 +23,24 @@ internal class ReviewContainerViewController: UIViewController, ContainerViewCon
...
-        contentController = ReviewViewController(imageData, success:
-            { imageData in
-                self.imageData = imageData
+        contentController = ReviewViewController(self.document!, successBlock:
+            { [unowned self] (document, shouldFinish) in
+                self.document = document
                 self.changes = true
-            }, failure: { error in
+                
+                if shouldFinish {
```
nsobadjiev: I really don't understand this `shouldFinish`...

GiniVision/Classes/ReviewViewController.swift
```diff
@@ -16,13 +16,27 @@ import UIKit
 public typealias ReviewSuccessBlock = (_ imageData: Data) -> ()
 
 /**
+ Block which will be executed each time the user rotates a picture. It contains the JPEG representation of the image including meta information about the rotated image. In the case of a PDF, it should proceed to analysis screen once it has been validated.
```
nsobadjiev: Block **that**

Also, why should clients validate the PDF? Shouldn't they only receive a valid one?

GiniVision/Classes/ReviewViewController.swift
```diff
@@ -16,13 +16,27 @@ import UIKit
...
+ 
+ - note: Component API only.
+ */
+public typealias ReviewScreenSuccessBlock = (_ document: GiniVisionDocument, _ shouldProceedToAnalysisScreen:Bool) -> ()
```
nsobadjiev: Still not convinced about that `shouldProceedToAnalysisScreen `...

GiniVision/Classes/ReviewViewController.swift
```diff
@@ -16,13 +16,27 @@ import UIKit
...
+
+/**
  Block which will be executed when an error occurs on the review screen. It contains a review specific error.
  
  - note: Component API only.
  */
 public typealias ReviewErrorBlock = (_ error: ReviewError) -> ()
 
 /**
+ Block which will be executed if an error occurs on the review screen.
```
nsobadjiev: Block that.

GiniVision/Classes/ReviewViewController.swift
```diff
@@ -62,35 +76,41 @@ public typealias ReviewErrorBlock = (_ error: ReviewError) -> ()
...
-        errorBlock = failure
+        self.successBlock = successBlock
+        self.failureBlock = failureBlock
+        
+        // Initialize currentDocument
+        currentDocument = document
         
         // Set meta information manager
-        metaInformationManager = ImageMetaInformationManager(imageData: imageData)
+        if document.type == .Image {
```
nsobadjiev: Having a "Document" metaphor is a cool idea. However, I'm still seeing these type checks all over the place. This defeats the purpose of hiding PDFs and images in a Document. I think you should rethink some of the architecture to make ti possible to not care what kind of document it is in places such as `ReviewViewController`.

Example/GiniVision.xcodeproj/project.pbxproj
```diff
@@ -358,13 +362,16 @@
 					};
 					607FACCF1AFB9204008FA782 = {
 						CreatedOnToolsVersion = 6.3.1;
-						DevelopmentTeam = JA825X8F7Z;
+						DevelopmentTeam = BQY9KX4V88;
```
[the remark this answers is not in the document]
kikettas: It is changed by Xcode automatically and I also don't know why it does it. I will try to figure out...

Example/GiniVision.xcodeproj/project.pbxproj
```diff
@@ -916,7 +925,7 @@
 				CODE_SIGN_ENTITLEMENTS = GiniVision_Example.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
-				DEVELOPMENT_TEAM = JA825X8F7Z;
+				DEVELOPMENT_TEAM = BQY9KX4V88;
```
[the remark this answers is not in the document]
kikettas: Commented above.

Example/GiniVision_Example.entitlements
kikettas: UIDocumentPickerViewController needs iCloud to be enabled, explained [here](https://developer.apple.com/documentation/uikit/uidocumentpickerviewcontroller).

Example/Tests/GINICameraViewControllerTests.swift
kikettas: Because otherwise there will be an ambiguous reference and Xcode does not know how to solve it.

Example/Tests/GiniVisionDocumentTests.swift
kikettas: I use to write tests adding those comments, because for me it is more readable and it helps me to split in different steps. But there is no problem if you want me to remove them.

GiniVision/Classes/CameraViewController.swift
kikettas: In order to keep the same methods and avoid companies to make any change to their current implementation, we added the new delegate method `didImport`. We use that property when we switch between calling that or method or `didCapture`.

GiniVision/Classes/CameraViewController.swift
kikettas: Yes, but we cannot change the ViewController name since it would involve more changes for current users. But I think that __Camera__ is referring to the Camera Screen not to the camera itself.

GiniVision/Classes/CameraViewController.swift
kikettas: It is not possible as I commented above.

GiniVision/Classes/CameraViewController.swift
kikettas: Do you think it is inconsistent? This is not a init parameter and I was forced to user "*Block" for ambiguity reasons.

GiniVision/Classes/CameraViewController.swift
kikettas: To be consistent with the other source configuration (camera).

GiniVision/Classes/Extensions.swift
kikettas: Because of the _UTI's_

GiniVision/Classes/GiniImageDocument.swift
kikettas: `NSItemProviderReading` needs to be implements by a class conforming `NSObjectProtocol`

GiniVision/Classes/ReviewViewController.swift
kikettas: As I explained before the code review, document validation should be also done on the ReviewViewController because the document could have been imported with _Open with..._.

GiniVision/Classes/ReviewViewController.swift
kikettas: I'm not convinced neither, but I haven't found a better solution since document validation is done also in ReviewViewController (otherwise we could go directly to analysis when the document is a PDF).

GiniVision/Classes/ReviewViewController.swift
kikettas: Actually I wanted to move the _ImageMetaInformationManager_ to the GiniImageDocument class, but since I have other things to do, I was waiting to this PR to be approved.

Example/Tests/GiniVisionDocumentTests.swift
kikettas: Actually It did not take much time on generating fake data, but I found a way to make it faster!
Thanks!

Example/Tests/GiniVisionDocumentTests.swift
kikettas: It could be useful 👌

Example/GiniVision/ComponentAPIAnalysisViewController.swift
kikettas: The idea would be injecting it on the constructor, but since we are using `Storyboards` it is not possible to use its initializer, is it?

Example/GiniVision/ComponentAPIReviewViewController.swift
kikettas: Me too, but as commented above I think that it is not possible since we are using `Storyboards`. Or is there other way?

GiniVision/Classes/ReviewContainerViewController.swift
kikettas: Yes, the name is too ambiguous. It should be `shouldProceedToAnalysisScreen`

GiniVision/Classes/GiniImageDocument.swift
```diff
@@ -18,7 +18,7 @@ final public class GiniImageDocument: NSObject, GiniVisionDocument {
     public var previewImage: UIImage?
     
     /**
-     Initializes a GiniImageDocument with a preview image (the image itself).
+     Initializes a GiniImageDocument.
      
      - Parameter data: PDF data
```
a-szotyori: Image data 😉

GiniVision/Classes/FilePickerManager.swift
```diff
@@ -57,9 +57,14 @@ extension FilePickerManager: UIDocumentPickerDelegate {
             let data = try Data(contentsOf: url)
             url.stopAccessingSecurityScopedResource()
             
-            if let document = GiniVisionDocumentFactory.create(withData: data) {
```
a-szotyori: I like the factory approach. I understand your consideration and I think you are avoiding a phantom requirement by removing the factory.

For the current state this approach is nice, readable and easy to understand. In case we introduce a new document type with different initializer parameters in the future we can alter this approach.

I would recommend keeping the factory, but moving the factory method to the GiniVisionDocument: `GiniVisionDocument.create(withData: data)`.

What do you think?

a-szotyori: 👍 Thanks for clarifying it: the `GiniImageDocument` will require an initializer with more parameters when you move the meta information manager to it.

Example/Tests/GINICameraViewControllerTests.swift
nsobadjiev: Wont compile, or code completion wouldn't work?

Example/GiniVision/ComponentAPIReviewViewController.swift
nsobadjiev: Well, you can start with an empty document. That's the alternative. But it's not ideal. Ok, lets keep it.

Example/GiniVision/ComponentAPIAnalysisViewController.swift
nsobadjiev: It's not, but still, I don't think having it force unwrapped is so good. Can you at least use "?"?

Example/GiniVision.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
nsobadjiev: That's fine, but I don't want to see such changes. Can you **not** commit them?
Just revert them before committing and creating the PR.

Example/Tests/GINICameraViewControllerTests.swift
kikettas: It won't compile. Since both parameters are `closures` in both methods, Xcode does not know what is the difference.

GiniVision/Classes/FilePickerManager.swift
kikettas: Apart from what we discussed earlier, I do not think that It would have been a good idea to put `create` factory method inside `GiniVisionDocument` protocol since It should not know about implementation details. I see protocols as 'a set of rules' that do not care about implementation details. What do you think?

kikettas:
Closes #50 and #52

Issue #55, opened 2017-09-01
File preview PDF (image render)


kikettas:
Each time a PDF is imported, an image from the first image should be shown on Review screen.

PR #62, opened 2017-09-15
Render preview image from PDF

kikettas:
When a PDF is imported, a preview image must be shown on analysis screen. 

### How to test
Try to import a PDF and then see that a preview image is shown.

### Merging
Automatic.

### Ticket 
Issue #55


Example/Tests/GINIPDFDocumentTests.swift
```diff
@@ -0,0 +1,58 @@
...
+    func testPDFNonRotated() {
+        let imageSize = pdfNonRotated.previewImage!.size
+        
+        XCTAssert(imageSize.height > imageSize.width, "Height must be greater than width since image has not rotationAngle and the pdf is in portrait orientation by default")
+    }
+    
+    func testPDFRotated90() {
+        let pdfDocument = GINIPDFDocumentTests.loadPDFDocument(withName: "testPDF-rotated90")
+        
+        let imageSize = pdfDocument.previewImage!.size
```
nsobadjiev: One note about using "!" in tests: It's true that crashing is not a big deal (and even sometimes welcome) in unit tests. However, if an exception is thrown in the test body, Xcode will mark it as "an unexpected failure". You will still see which test failed, but the error message might not be as good (for instance, you wont see the beautiful assert message below). That being said, I still wouldn't consider using "!" here an error. But if it is easy to avoid it, you might reconsider.

Example/Tests/GINIPDFDocumentTests.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+
+import XCTest
+@testable import GiniVision
+
+class GINIPDFDocumentTests: XCTestCase {
+  
+    let pdfNonRotated = GINIPDFDocumentTests.loadPDFDocument(withName: "testPDF")
+    
+    func testNumberPages() {
```
nsobadjiev: Testing with a PDF with more than one page might be interesting as a regression test. If you can find a two page one, you should add another test to make sure the numberPages property really counts pages.

GiniVision/Classes/GiniPDFDocument.swift
```diff
@@ -29,10 +29,11 @@ final public class GiniPDFDocument: NSObject, GiniVisionDocument {
...
         self.data = data
-        self.previewImage = UIImageNamedPreferred(named: "cameraDefaultDocumentImage") // Here should be the first rendered page
-        if let dataProvider = CGDataProvider(data: data as CFData) {
-            let pdfDocument = CGPDFDocument(dataProvider)
-            self.numberPages = pdfDocument?.numberOfPages ?? 0
+        super.init()
+        
+        if let dataProvider = CGDataProvider(data: data as CFData), let pdfDocument = CGPDFDocument(dataProvider) {
+            self.numberPages = pdfDocument.numberOfPages
+            self.previewImage = renderFirstPage(fromPdf: pdfDocument)
```
nsobadjiev: Can this be "lazy"? You don't need the preview right from the initialise.

GiniVision/Classes/GiniPDFDocument.swift
```diff
@@ -56,6 +57,42 @@ final public class GiniPDFDocument: NSObject, GiniVisionDocument {
         }
     }
     
+    fileprivate func renderFirstPage(fromPdf pdf:CGPDFDocument) -> UIImage? {
+        var pdfImage:UIImage?
+        let pdfDoc = pdf
+        
+        if let pdfPage:CGPDFPage = pdfDoc.page(at: 1) {
```
nsobadjiev: I hope this 1 is safe and wont crash for PDF's with 0 zero pages (if that's possible)...

GiniVision/Classes/GiniPDFDocument.swift
```diff
@@ -56,6 +57,42 @@ final public class GiniPDFDocument: NSObject, GiniVisionDocument {
...
+            var pageRect:CGRect = pdfPage.getBoxRect(.cropBox)
+            
+            // In case that the image was rotated 90 or 270, final rect should be rotated to portrait             
+            if pdfPage.rotationAngle == 90 || pdfPage.rotationAngle == 270 {
```
nsobadjiev: I would extract that in a method like `normalizeOrientation` or something...

GiniVision/Classes/GiniPDFDocument.swift
```diff
@@ -56,6 +57,42 @@ final public class GiniPDFDocument: NSObject, GiniVisionDocument {
...
+                let tempWidth = pageRect.size.width
+                pageRect.size.width = pageRect.size.height
+                pageRect.size.height = tempWidth
+            }
+            
+            // Create context
```
nsobadjiev: Useless comment, IMHO

GiniVision/Classes/GiniPDFDocument.swift
```diff
@@ -56,6 +57,42 @@ final public class GiniPDFDocument: NSObject, GiniVisionDocument {
...
+            UIGraphicsBeginImageContextWithOptions(CGSize(width: pageRect.width, height: pageRect.height), false, 0.0)
+            let context:CGContext = UIGraphicsGetCurrentContext()!
+            
+            // Fill context color
```
nsobadjiev: Also useless. You don't have to copy all the comments from SO :P

GiniVision/Classes/GiniPDFDocument.swift
```diff
@@ -56,6 +57,42 @@ final public class GiniPDFDocument: NSObject, GiniVisionDocument {
...
+            context.setFillColor(UIColor.white.cgColor)
+            context.fill(pageRect)
+            
+            // Align PDF's cropBox to the context
```
nsobadjiev: This one would be the only comment here that would be useful. However, it's vague and I still don't understand it.

GiniVision/Classes/ReviewViewController.swift
```diff
@@ -193,6 +193,14 @@ public typealias ReviewScreenFailureBlock = (_ error: GiniVisionError) -> ()
         updateMinZoomScaleForSize(scrollView.bounds.size)
     }
     
+    public override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        
+        // On initialization imageView frame is (0,0) so the image needs to be centered
+        // inside the ScrollView when its size has changed
+        self.updateConstraintsForSize(scrollView.bounds.size)
```
nsobadjiev: If it's about the view changing size, wouldn't a method like `didLayoutSubviews` be more suitable?

Also, "// On initialization imageView'**S** frame"

GiniVision/Classes/GiniPDFDocument.swift
kikettas: I could look like copied from SO, but actually they were checks for me since I was struggling with an issue related to image orientation.
For me they were useless at all, but if you want they could be removed.

GiniVision/Classes/GiniPDFDocument.swift
kikettas: Commented above

GiniVision/Classes/GiniPDFDocument.swift
kikettas: I think it depends, because for instance when you use "Open with..." method you need the preview image asap. So I would keep this here (also it is useful since CGPDFDocument is used for the number of pages and it can be used to render the first image)

GiniVision/Classes/GiniPDFDocument.swift
kikettas: It would be `nil` and it won't continue, the only thing is that the pdfImage would be `nil`

GiniVision/Classes/GiniPDFDocument.swift
kikettas: I like the idea! 👌

Example/Tests/GINIPDFDocumentTests.swift
kikettas: Yes, that could be another test case! 👍

GiniVision/Classes/ReviewViewController.swift
kikettas: You are right, I don't know why I decided to put it on the viewWillAppear method. Thank you! 👍

GiniVision/Classes/GiniPDFDocument.swift
```diff
@@ -56,6 +57,48 @@ final public class GiniPDFDocument: NSObject, GiniVisionDocument {
...
+            context.translateBy(x: 0, y: pageRect.size.height)
+            context.scaleBy(x: 1, y: -1)
+            context.concatenate(pdfPage.getDrawingTransform(.cropBox, rect: pageRect, rotate: 0, preserveAspectRatio: true))
```
a-szotyori: Will that infamous pdf (with it's page rotated 270°) be rendered correctly?

kikettas: Yes, It will work as expected (different orientations will also work). ☺️

Issue #58, opened 2017-09-08
Add Image Source to Exif User Comment


a-szotyori:
To be able to determine where the image comes from we add the `Source` and `ImportMethod` keys to the user comment.

### `Source`

Contains the source of the image: camera or an external app.

#### Values:
- `camera`
- `external` if source app info was not available
- name of the source app: CameraRoll, Photos, Documents, etc

### `ImportMethod`

Used Only when the source is an external app. Contains information about how the image was imported: open with, or picker

#### Values:
- `openwith` when document was imported by opening GVL from another app
- `picker` when document was imported from the picker in the camera screen

PR #63, opened 2017-09-19
Uncouple image meta information manager from views

kikettas:
`ImageMetaInformationManager` is coupled to the views and they should not know anything about the manager.

### How to test
Take a picture or import an image an see that all meta information is added to it. Also rotate the picture and check that `RotDeltaDeg` has changed.

### Merging
Automatic.

### Ticket 
Issue #58

### DIFF
--- a/GiniVision/Classes/GiniImageDocument.swift
+++ b/GiniVision/Classes/GiniImageDocument.swift
@@ -17,16 +17,23 @@ final public class GiniImageDocument: NSObject, GiniVisionDocument {
     public var data:Data
     public var previewImage: UIImage?
     
+    fileprivate let metaInformationManager:ImageMetaInformationManager
+    
     /**
      Initializes a GiniImageDocument.
      
      - Parameter data: PDF data
+     - Parameter deviceOrientation: Device orientation when a picture was taken from the camera. In other cases it should be `nil`
      
      */
     
-    public init(data: Data) {
-        self.data = data
+    public init(data: Data, deviceOrientation:UIInterfaceOrientation? = nil) {
         self.previewImage = UIImage(data: data)
+        self.metaInformationManager = ImageMetaInformationManager(imageData: data, deviceOrientation:deviceOrientation)
+        self.metaInformationManager.filterMetaInformation()
+        
+        // Add meta information to image document data
+        self.data = metaInformationManager.imageData() ?? data
     }
     
     /**
@@ -44,6 +51,12 @@ final public class GiniImageDocument: NSObject, GiniVisionDocument {
         } else {
             throw DocumentValidationError.fileFormatNotValid
         }
+    }
+    
+    func rotateImage(degrees:Int, imageOrientation:UIImageOrientation) {
+        metaInformationManager.rotate(degrees: 90, imageOrientation: imageOrientation)
+        guard let data = metaInformationManager.imageData() else { return }
+        self.data = data
     }
 }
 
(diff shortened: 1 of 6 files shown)


GiniVision/Classes/GiniImageDocument.swift
```diff
@@ -17,16 +17,23 @@ final public class GiniImageDocument: NSObject, GiniVisionDocument {
...
+     - Parameter deviceOrientation: Device orientation when a picture was taken from the camera. In other cases it should be `nil`
      
      */
     
-    public init(data: Data) {
-        self.data = data
+    public init(data: Data, deviceOrientation:UIInterfaceOrientation? = nil) {
         self.previewImage = UIImage(data: data)
+        self.metaInformationManager = ImageMetaInformationManager(imageData: data, deviceOrientation:deviceOrientation)
+        self.metaInformationManager.filterMetaInformation()
```
nsobadjiev: I know this method was there before you, but I think it's misleading. First of all, it doesn't filter anything - just adds the meta tags. Second of all, I'd remove it from the public interface altogether. You will always want the tags, so just call it internally at the end of the init.

GiniVision/Classes/ImageMetaInformationManager.swift
```diff
@@ -71,8 +71,20 @@ internal class ImageMetaInformationManager {
         }
     }
     
-    func imageData(withCompression compression: CGFloat = JPEGDefaultCompression) -> Data? {
-        return generateImage(withMetaInformation: metaInformation, andCompression: compression)
+    // Returns the current image but with all meta information added
+    func addMetaInformationToImage(withCompression compression: CGFloat = JPEGDefaultCompression) -> Data? {
```
nsobadjiev: This name is misleading. it makes me think it will add the meta info to the existing image, but it actually returns a new image data. A more swift-like name would be `imageByAddingMetadata` or something.

kikettas: Yes, the name you suggest is more swift-like. 👍

GiniVision/Classes/GiniImageDocument.swift
kikettas: It makes sense, since it is only used when the manager is initialized. But, since this class needs to be refactored, I think we should keep the method name until we refactor it in order to avoid misunderstandings in the future.

GiniVision/Classes/GiniImageDocument.swift
```diff
@@ -17,16 +17,23 @@ final public class GiniImageDocument: NSObject, GiniVisionDocument {
...
+        
+        // Add meta information to image document data
+        self.data = metaInformationManager.addMetaInformationToImage() ?? data
```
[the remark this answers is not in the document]
kikettas: I came up with an idea. I added an assert when the optional is unwrapped, so it will fail when the  it could not add metadata to an image.

GiniVision/Classes/ImageMetaInformationManager.swift
```diff
@@ -71,8 +71,20 @@ internal class ImageMetaInformationManager {
...
+        guard let image = imageData else { return nil }
+        guard let information = metaInformation else { return nil }
+        
+        let targetData = NSMutableData()
+        
+        let destination = CGImageDestinationCreateWithData(targetData, kUTTypeJPEG, 1, nil)!
+        let source = CGImageSourceCreateWithData(image as CFData, nil)
+        information.setValue(compression, forKey: kCGImageDestinationLossyCompressionQuality as String)
+        
+        CGImageDestinationAddImageFromSource(destination, source!, 0, information as CFDictionary)
```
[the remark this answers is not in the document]
kikettas: Since the method returns `nil` it wasn't a big deal 😉

Issue #64, opened 2017-09-19
Remove document validation from ReviewScreen


kikettas:
When a document is imported with "Open with" it should be validated before loading both ScreenAPI or ComponentAPI, not in ReviewScreen.

PR #74, opened 2017-09-22
Remove document validation on review screen

kikettas:
Some documents are not reviewable, so they do not have to go through the review screen. On the Component API 

### How to test
Import a PDF and see that it doesn't go through the review screen. If instead you import an image or take a picture with the camera, it should be reviewable from review screen. 
Also using the "Open with" method, you should have the same result.

### Merging
Automatic.

### Ticket 
Issue #64

### DIFF
--- a/Example/GiniVision/AppDelegate.swift
+++ b/Example/GiniVision/AppDelegate.swift
@@ -41,7 +41,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, open url: URL, sourceApplication: String?, annotation: Any) -> Bool {
         // This is only a shortcut for demo purposes since if the current root view controller is not
         // the ScreenAPIViewController (a GiniVisionDelegate), it won't do anything.
-        guard let navVC = window?.rootViewController as? UINavigationController, let screenAPIVC = navVC.viewControllers.first as? ScreenAPIViewController else {
+        guard let navVC = window?.rootViewController as? UINavigationController, let selectAPIVC = navVC.viewControllers.first as? SelectAPIViewController else {
             return false
         }
         
@@ -58,16 +58,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         do {
             try document?.validate()
             
-            // 4. Create alert which allows user to open imported file either with ScreenAPI or ComponentAPI
+            // 4. Create an alert which allow users to open imported file either with the ScreenAPI or the ComponentAPI
             alertViewController = UIAlertController(title: "Importierte Datei", message: "Möchten Sie die importierte Datei mit dem ScreenAPI oder ComponentAPI verwenden?", preferredStyle: .alert)
             
             alertViewController.addAction(UIAlertAction(title: "Screen API", style: .default) { _ in
-                screenAPIVC.present(screenAPIVC.giniScreenAPI(withImportedDocument: document), animated: true, completion: nil)
+                selectAPIVC.present(selectAPIVC.giniScreenAPI(withImportedDocument: document), animated: true, completion: nil)
             })
             
             alertViewController.addAction(UIAlertAction(title: "Component API", style: .default) { _ in
-                if let componentAPI = screenAPIVC.giniComponentAPI(withImportedDocument: document) {
-                    screenAPIVC.present(componentAPI, animated: true, completion: nil)
+                if let componentAPI = selectAPIVC.giniComponentAPI(withImportedDocument: document) {
+                    selectAPIVC.present(componentAPI, animated: true, completion: nil)
                 }
             })
         } catch {
@@ -79,7 +79,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
         
         // 5. Present alert
-        screenAPIVC.present(alertViewController, animated: true, completion: nil)
+        selectAPIVC.present(alertViewController, animated: true, completion: nil)
         
         return true
     }
--- a/Example/GiniVision/AppDelegate.swift
+++ b/Example/GiniVision/AppDelegate.swift
@@ -66,9 +66,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             })
             
             alertViewController.addAction(UIAlertAction(title: "Component API", style: .default) { _ in
-                if let componentAPI = selectAPIVC.giniComponentAPI(withImportedDocument: document) {
-                    selectAPIVC.present(componentAPI, animated: true, completion: nil)
-                }
+//                if let componentAPI = selectAPIVC.giniComponentAPI(withImportedDocument: document) {
+//                    selectAPIVC.present(componentAPI, animated: true, completion: nil)
+//                }
+                let componentAPICoordinator = ComponentAPICoordinator(document: document)
+                componentAPICoordinator.start(from: selectAPIVC)
             })
         } catch {
             // 4.1. Create alert which shows an error pointing out that it is not a valid document
--- a/Example/GiniVision/AppDelegate.swift
+++ b/Example/GiniVision/AppDelegate.swift
@@ -34,7 +34,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         self.giniSDK = builder?.build()
         
         print("Gini Vision Library for iOS (\(GiniVision.versionString)) / Client id: \(clientId)")
-        
+
         return true
     }
     
--- a/Example/GiniVision/AppDelegate.swift
+++ b/Example/GiniVision/AppDelegate.swift
@@ -34,7 +34,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         self.giniSDK = builder?.build()
         
         print("Gini Vision Library for iOS (\(GiniVision.versionString)) / Client id: \(clientId)")
-
+        
         return true
     }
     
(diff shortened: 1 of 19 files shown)


Example/GiniVision/NoResultViewController.swift
```diff
@@ -19,6 +19,10 @@ class NoResultViewController: UIViewController {
     }
     
     @IBAction func retry(_ sender: AnyObject) {
-        _ = self.navigationController?.popToRootViewController(animated: true)
+        if let navVC = navigationController, !isFirstViewController(inNavController: navVC) {
```
nsobadjiev: Another example of a view controller that knows too much about how it is presented. Although, here I might let it slide because at least it will work in most of the types of presentation.

Example/GiniVision/ResultTableViewController.swift
```diff
@@ -32,12 +32,16 @@ class ResultTableViewController: UITableViewController {
...
         super.viewDidLoad()
-        
+
         // If a valid document is set, send feedback on it.
         // This is just to show case how to give feedback using the Gini SDK for iOS.
         // In a real world application feedback should be triggered after the user has evaluated and eventually corrected the extractions.
         sendFeedback(forDocument: document)
     }
+        
+    func closeAction() {
```
nsobadjiev: Different class, same close action....

GiniVision/Classes/GiniVision.swift
```diff
@@ -93,17 +93,20 @@ import UIKit
...
      */
-    public class func viewController(withDelegate delegate: GiniVisionDelegate, importedFile:Data? = nil) -> UIViewController {
+    public class func viewController(withDelegate delegate: GiniVisionDelegate, importedDocument:GiniVisionDocument? = nil) -> UIViewController {
         let viewController:UIViewController
-        let documentBuilder = GiniVisionDocumentBuilder(data:importedFile, documentSource: .external)
-        documentBuilder.importMethod = .openWith
         
-        if let document = documentBuilder.build() {
-            viewController = ReviewContainerViewController(document: document)
+        if let document = importedDocument {
```
nsobadjiev: Saw exactly the same code in CameraContainerViewController. Code duplication...

Example/GiniVision/AppDelegate.swift
```diff
@@ -48,21 +48,37 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            
+            // 4. Create alert which allows user to open imported file either with ScreenAPI or ComponentAPI
+            alertViewController = UIAlertController(title: "Importierte Datei", message: "Möchten Sie die importierte Datei mit dem ScreenAPI oder ComponentAPI verwenden?", preferredStyle: .alert)
+            
+            alertViewController.addAction(UIAlertAction(title: "Screen API", style: .default) { _ in
+                screenAPIVC.present(screenAPIVC.giniScreenAPI(withImportedDocument: document), animated: true, completion: nil)
+            })
+            
+            alertViewController.addAction(UIAlertAction(title: "Component API", style: .default) { _ in
+                if let componentAPI = screenAPIVC.giniComponentAPI(withImportedDocument: document) {
```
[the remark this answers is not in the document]
kikettas: Yes, the name of the __main__ view controller could lead to misunderstandings. Needs to be renamed. 👍

Example/GiniVision/AppDelegate.swift
```diff
@@ -58,17 +58,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
             alertViewController.addAction(UIAlertAction(title: "Screen API", style: .default) { _ in
-                screenAPIVC.present(screenAPIVC.giniScreenAPI(withImportedDocument: document), animated: true, completion: nil)
+                selectAPIVC.present(selectAPIVC.giniScreenAPI(withImportedDocument: document), animated: true, completion: nil)
             })
             
             alertViewController.addAction(UIAlertAction(title: "Component API", style: .default) { _ in
-                if let componentAPI = screenAPIVC.giniComponentAPI(withImportedDocument: document) {
-                    screenAPIVC.present(componentAPI, animated: true, completion: nil)
-                }
+//                if let componentAPI = selectAPIVC.giniComponentAPI(withImportedDocument: document) {
```
nsobadjiev: I suppose this will never be uncommented. Just remove it.

Example/GiniVision/ComponentAPIAnalysisViewController.swift
```diff
@@ -10,6 +10,13 @@ import UIKit
 import GiniVision
 import Gini_iOS_SDK
 
+protocol ComponentAPIAnalysisScreenDelegate:class {
+    func didCancelAnalysis()
```
nsobadjiev: Just a question - why don't you use the conventional Cocoa framework delegate style? Here, it would have been:

`analysisScreenDidCancelAnalysis(_ screen: ComponentAPIAnalysisViewController)`

Example/GiniVision/ComponentAPIAnalysisViewController.swift
```diff
@@ -42,60 +50,27 @@ class ComponentAPIAnalysisViewController: UIViewController {
...
-        NotificationCenter.default.removeObserver(self)
-    }
-    
-    override func didMove(toParentViewController parent: UIViewController?) {
-        super.didMove(toParentViewController: parent)
-        
-        // Cancel analysis process to avoid unnecessary network calls.
-        if parent == nil {
-            AnalysisManager.sharedManager.cancelAnalysis()
+        if isMovingFromParentViewController || isBeingDismissed {
```
nsobadjiev: It's nice that you removed the didMove method! Are you sure the viewDidDisappear is enough? It will be called even when a new view controller is pushed on top.

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -0,0 +1,269 @@
...
+import GiniVision
+import Gini_iOS_SDK
+
+final class ComponentAPICoordinator {
+    
+    fileprivate var document:GiniVisionDocument?
+    fileprivate var navigationController: UINavigationController?
+    fileprivate var tabBarController: UITabBarController?
+    fileprivate var analysisScreen: ComponentAPIAnalysisViewController? {
+        return navigationController?.childViewControllers.flatMap { $0 as? ComponentAPIAnalysisViewController }.first
```
nsobadjiev: I'm not convinced about that. Can't you hold a normal reference to the view controller. For instance, when you present it, you can keep a reference to it.

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -0,0 +1,269 @@
...
+    }
+    fileprivate var resultsScreen: ResultTableViewController? {
+        return navigationController?.childViewControllers.flatMap { $0 as? ResultTableViewController }.first
```
nsobadjiev: Same as above.

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -0,0 +1,269 @@
...
+    }
+    
+    init(document:GiniVisionDocument?){
+        self.document = document
+        
+        // 1. Create and set a custom configuration object needs to be done once before using any component of the Component API.
```
nsobadjiev: A lonely "1"... without a "2".
I think the whole comment is out of date.

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -0,0 +1,269 @@
...
+                showCameraScreen()
+            }
+            
+            rootViewController.present(tabBar, animated: true, completion: nil)
+        }
+    }
+    
+    // MARK: Show screens
+    fileprivate func showCameraScreen() {
+        let cameraContainer = UIStoryboard(name: "Main", bundle: nil).instantiateViewController(withIdentifier: "ComponentAPICamera") as! ComponentAPICameraViewController
```
nsobadjiev: You are using this code all the time. Maybe you can extract it somewhere.

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -0,0 +1,269 @@
...
+        let hasPayFive = result.filter { payFive.contains($0.0) }.count > 0
+        
+        if hasPayFive {
+            showResultsTableScreen(forDocument: document, withResult: result)
+        } else {
+            showNoResultsScreen()
+        }
+        
+        if let analysisScreen = analysisScreen {
+            removeFromStack(analysisScreen)
```
nsobadjiev: Wont it always be the top view controller?

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -0,0 +1,269 @@
...
+        dismissTabBarController()
+    }
+}
+
+// MARK: ComponentAPIReviewScreenDelegate
+
+extension ComponentAPICoordinator: ComponentAPIReviewScreenDelegate {
+    
+    func didReview(documentReviewed: GiniVisionDocument) {
+        // Analyze reviewed data because changes were made by the user during review.
```
nsobadjiev: They code is fairly self-explanatory IMHO. You don't need the comment.

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -0,0 +1,269 @@
...
+        }
+        
+        // Restart analysis if it was canceled and is currently not running.
+        if !AnalysisManager.sharedManager.isAnalyzing {
+            if let documentData = document?.data {
+                AnalysisManager.sharedManager.analyzeDocument(withData: documentData, cancelationToken: CancelationToken(), completion: nil)
+            }
+        }
+        
+        // Show analysis screen if no results are in yet and no changes were made.
```
nsobadjiev: Same thing. A bit redundant.

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -0,0 +1,269 @@
...
+            showResultsTableScreen(forDocument: document, withResult: result)
+        } else {
+            showNoResultsScreen()
+        }
+        
+        if let analysisScreen = analysisScreen {
+            removeFromStack(analysisScreen)
+        }
+        
+        if navigationController?.viewControllers.count == 2 {
```
nsobadjiev: Again, this view controller counting thing... I hoped I'd never see it again.

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -0,0 +1,269 @@
...
+            if let resultsScreen = resultsScreen {
+                resultsScreen.navigationItem.rightBarButtonItem = UIBarButtonItem(title: "Schließen", style: .plain, target: self, action: #selector(dismissTabBarController))
+            }
+        }
+    }
+    
+    fileprivate func handleExistinResults() {
```
nsobadjiev: Existin?

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -0,0 +1,269 @@
...
+    
+    func didCancelAnalysis() {
+        AnalysisManager.sharedManager.cancelAnalysis()
+    }
+    
+    func didAppear() {
+        // Subscribe to analysis events which will be fired when the analysis process ends.
+        // Either with a valid result or an error.
+        NotificationCenter.default.addObserver(self, selector: #selector(handleAnalysis(errorNotification:)), name: NSNotification.Name(rawValue: GINIAnalysisManagerDidReceiveErrorNotification), object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(handleAnalysis(resultNotification:)), name: NSNotification.Name(rawValue: GINIAnalysisManagerDidReceiveResultNotification), object: nil)
```
nsobadjiev: I hate using notifications for things like this. But I guess it was already like so. Carry on...

Example/GiniVision/AppDelegate.swift
kikettas: It will never be. Removed!

Example/GiniVision/ComponentAPICoordinator.swift
kikettas: Yes, I hate this too but I think it could be removed in a future.

Example/GiniVision/ComponentAPIAnalysisViewController.swift
kikettas: I don't know, I see it simpler but yes, I should use Cocoa framework delegate style.

Example/GiniVision/ComponentAPICoordinator.swift
kikettas: Since it is a small collection of view controllers (4 at most) I found it a good approach (flatMap does not have a good performance with big collections).
But yes, it could be confusing and less readable.

Example/GiniVision/ComponentAPICoordinator.swift
kikettas: Answered above

Example/GiniVision/ComponentAPIAnalysisViewController.swift
kikettas: Yes, but since it is the last view controller before the results, it won't be a problem.

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -0,0 +1,269 @@
...
+        
+        // 1. Create and set a custom configuration object needs to be done once before using any component of the Component API.
+        let giniConfiguration = GiniConfiguration()
+        giniConfiguration.debugModeOn = true
+        GiniVision.setConfiguration(giniConfiguration)
+    }
+    
+    func start(from rootViewController:UIViewController) {
+        if let tabBar = UIStoryboard(name: "Main", bundle: nil).instantiateViewController(withIdentifier: "ComponentAPI") as? UITabBarController,
+            let navBar = tabBar.viewControllers?.first as? UINavigationController {
```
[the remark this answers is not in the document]
kikettas: Yes, they both could be created on the coordinator and instantiate the other tab also in the coordinator.

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -0,0 +1,269 @@
...
+        tabBarController?.dismiss(animated: true, completion: nil)
+    }
+    
+    fileprivate func addCloseButtonIfNeeded(onViewController viewController: UIViewController) {
+        if let navBar = navigationController, navBar.viewControllers.isEmpty {
+            viewController.navigationItem.leftBarButtonItem = UIBarButtonItem(title: "Schließen", style: .plain, target: self, action: #selector(dismissTabBarController))
+        }
+    }
+    
+    fileprivate func removeFromStack(_ viewController: UIViewController) {
```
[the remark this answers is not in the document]
kikettas: I don't like neither this method and the one where it is called. In a future refactor it will be removed for sure!

Example/GiniVision/ComponentAPICoordinator.swift
kikettas: I think that it was made thinking only on the original flow, but as I said before, this method will be removed in a future refactor.

Example/GiniVision/ComponentAPICoordinator.swift
kikettas: Explained the reason above.

Issue #59, opened 2017-09-11
Import file (PDF, image) feature toggle


kikettas:
It should be possible to enable/disable the import file feature using a toggle

PR #75, opened 2017-09-25
File import toggle

kikettas:
File import feature should be enabled through `GiniConfiguration`, being disabled by default.

### How to test
Set `enabledFileImport` to `true` and see that file import feature is enabled. Then set it to `false` and see that it is disabled.

### Merging
Automatic.

### Ticket 
Issue #59

### DIFF
--- a/GiniVision/Classes/GiniConfiguration.swift
+++ b/GiniVision/Classes/GiniConfiguration.swift
@@ -109,6 +109,13 @@ import UIKit
     
     
     // MARK: Camera options
+    
+    /**
+     Toggles file import feature.
+
+     */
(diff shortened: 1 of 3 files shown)


GiniVision/Classes/GiniConfiguration.swift
```diff
@@ -109,6 +109,13 @@ import UIKit
     
     
     // MARK: Camera options
+    
+    /**
+     Toggles file import feature.
```
nsobadjiev: Toggles *the* import feature.
Also, only 4 words might not be enough for people to understand the parameter completely. And since the configuration is user facing, it might make sense to explain it a little better.

PR #76, opened 2017-09-26
File types supported toggle for file import feature

kikettas:
File types supported for file import feature. It should be specified through `GiniConfiguration`, being `none` by default.

### How to test
Set `fileImportSupportedTypes` to `.all` and see that file import feature is enabled for all types. If instead `.pdf` is set, the app will only handle PDF files (not pictures).
In case that none is specified, it should handle all file types (pictures and pdfs)

### Merging
Automatic.

### Ticket 
Issue #59

### DIFF
--- a/GiniVision/Classes/CameraViewController.swift
+++ b/GiniVision/Classes/CameraViewController.swift
@@ -356,7 +356,7 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> ()
         let alertViewController = UIAlertController(title: nil, message: nil, preferredStyle: .actionSheet)
         var alertViewControllerMessage = "Dokumente importieren"
         
-        if GiniConfiguration.sharedConfiguration.fileImportSupportedTypes == .all {
+        if GiniConfiguration.sharedConfiguration.fileImportSupportedTypes == .pdf_and_images {
             alertViewController.addAction(UIAlertAction(title: "Camera roll", style: .default) { [unowned self] _ in
                 self.filePickerManager.showGalleryPicker(from: self)
             })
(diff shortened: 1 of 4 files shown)


GiniVision/Classes/CameraViewController.swift
```diff
@@ -354,18 +354,24 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> ()
...
+        
+        if GiniConfiguration.sharedConfiguration.fileImportSupportedTypes == .all {
+            alertViewController.addAction(UIAlertAction(title: "Camera roll", style: .default) { [unowned self] _ in
+                self.filePickerManager.showGalleryPicker(from: self)
+            })
+            alertViewControllerMessage = "Fotos oder Dokumente importieren"
+        }
         
-        alertViewController.addAction(UIAlertAction(title: "Documents", style: .default) { [unowned self] _ in
+        alertViewController.addAction(UIAlertAction(title: "Dokumente", style: .default) { [unowned self] _ in
```
kikettas: This strings are changed in other branch, so don't pay attention to them

GiniVision/Classes/GiniConfiguration.swift
```diff
@@ -29,6 +29,11 @@ import UIKit
         return sharedConfiguration.debugModeOn
     }
     
+    public enum GiniVisionImportFileTypes {
+        case none
+        case pdf
+        case all
```
a-szotyori: I would rename `all` to `pdf_and_images`. To convey which types are enabled.

nsobadjiev:
Why is this merged without my review???

Issue #61, opened 2017-09-13
Exif handling for non-jpeg images


a-szotyori:
Since png and gif are also supported we have to implement a solution for these regarding exif (meta information) handling.

PR #77, opened 2017-09-26
Exif handle for GIFs and PNGs.

kikettas:
GIFs and PNGs images should not have meta information and should not be converted to JPEG. 

### How to test
Import a PNG or GIF image and see that its bytes count is the same when the image is analyzed.

### Merging
Automatic.

### Ticket 
Issue #61

### DIFF
--- a/GiniVision/Classes/GiniImageDocument.swift
+++ b/GiniVision/Classes/GiniImageDocument.swift
@@ -35,7 +35,6 @@ final public class GiniImageDocument: NSObject, GiniVisionDocument {
             self.data = dataWithMetadata
         } else {
             self.data = data
-            assertionFailure("It wasn't possible to add metadata to the image")
         }
     }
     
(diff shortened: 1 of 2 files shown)


GiniVision/Classes/GiniImageDocument.swift
```diff
@@ -58,10 +57,7 @@ final public class GiniImageDocument: NSObject, GiniVisionDocument {
     
     func rotateImage(degrees:Int, imageOrientation:UIImageOrientation) {
         metaInformationManager.rotate(degrees: 90, imageOrientation: imageOrientation)
-        guard let data = metaInformationManager.imageByAddingMetadata() else {
-            assertionFailure("It wasn't possible to add metadata to the image")
```
nsobadjiev: Why no assertion? Now it's even more likely that the call fails.

kikettas: Because PNG and GIF could be rotated even it doesn't have any effect in them.

PR #78, opened 2017-10-05
Show photo library permissions denied dialog

kikettas:
When photo library permission is denied, it should appear a dialog giving the chance to grant it on iOS settings.

### How to test
Reinstall app, choose _Fotos_ option on file import dialog and deny permissions. Then try again to open the Camera roll and see that the dialog appears.

### Merging
Automatic.

### Ticket 
Issue #65

### DIFF
--- a/GiniVision/Classes/CameraContainerViewController.swift
+++ b/GiniVision/Classes/CameraContainerViewController.swift
@@ -53,7 +53,7 @@ internal class CameraContainerViewController: UIViewController, ContainerViewCon
             switch error {
             case CameraError.notAuthorizedToUseDevice:
                 print("GiniVision: Camera authorization denied.")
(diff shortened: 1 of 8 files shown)


GiniVision/Classes/CameraContainerViewController.swift
```diff
@@ -165,5 +168,29 @@ internal class CameraContainerViewController: UIViewController, ContainerViewCon
         
         self.contentController.present(alertViewController, animated: true, completion: nil)
     }
-
+    
+    fileprivate func showPhotoLibraryPermissionDeniedError() {
```
a-szotyori: We should show this alert in the `CameraViewController` ourselves so that it is shown using the Component API, too. Clients can configure the message , but we take care of the permission alert inside the library => no public API changes required.

Issue #72, opened 2017-09-20
Tip popover for document import button


a-szotyori:
On first launch of the new version a popover should be shown telling users about the new document import feature.

UI:
- phone: https://projects.invisionapp.com/d/main#/projects/prototypes/12011059
- tablet: https://projects.invisionapp.com/d/main#/projects/prototypes/12036352

PR #81, opened 2017-10-05
Tip popover for document import button

kikettas:
When the app is opened for the first time, it shows a tip next to the file import button. It shouldn't appear again once it has been shown.

### How to test
Open the example app, go to the camera screen (either Screen or Component API) and see that the tip appears next to file import button. If you open the camera screen again you won't see it.

### Merging
Automatic.

### Ticket 
Issue #72


GiniVision/Classes/ToolTipView.swift
```diff
@@ -0,0 +1,303 @@
...
+    fileprivate var maxWidthOnIpad:CGFloat = 375
+    fileprivate var padding:(top:CGFloat, left:CGFloat, right: CGFloat, bottom: CGFloat) = (16, 16, 16, 16)
+    
+    fileprivate var textWidth:CGFloat {
+        guard let superview = superview else { return 0 }
+        if UIDevice.current.isIpad {
+            return maxWidthOnIpad - padding.left - padding.right - margin.left - margin.right - closeButtonWidth - itemSeparation
+        } else {
+            return superview.frame.width - padding.left - padding.right - margin.left - margin.right - closeButtonWidth - itemSeparation
+        }
```
a-szotyori: Could be simplified and clarified a bit:
```
let maxWidth = UIDevice.current.isIpad ? maxWidthOnIpad : superview.frame.width
let padding = padding.left + padding.right
let margin = margin.left + margin.right
return maxWidth - padding - margin - closeButtonWidth - itemSeparation
```

GiniVision/Classes/ToolTipView.swift
```diff
@@ -0,0 +1,303 @@
...
+        let frameHeight = max(textSize.height, closeButtonHeight) + padding.top + padding.bottom + margin.top + margin.bottom
+        let frameWidth = textSize.width + closeButtonWidth + padding.left + padding.right + itemSeparation + margin.left + margin.right
+        let size = CGSize(width: frameWidth, height: frameHeight)
+        
+        var x:CGFloat = 0
+        var y:CGFloat = 0
+        switch toolTipPosition {
+        case .above:
+            if referenceViewAbsoluteFrame.origin.y - size.height < 0 {
+                assertionFailure("The tip cannot be shown outside the super view")
```
a-szotyori: Could these assertions lead to crashing the app? Can these be caught? Would be bad if our lib crashes apps when for some reason the super view's size is wrong.

As an alternative we could fail gracefully and use alternative positions if the desired one cannot be used.

GiniVision/Classes/ToolTipView.swift
```diff
@@ -0,0 +1,303 @@
...
+    }
+    
+    fileprivate func absoluteFrame(for view:UIView, inside superView:UIView?) -> CGRect? {
+        guard let superView = superView, let referenceViewParent = referenceView.superview else { return nil }
+        
+        return referenceViewParent.convert(referenceView.frame, to: superView)
+    }
+    
+    // MARK: Constraints
+    fileprivate func setupConstraints() {
```
a-szotyori: OMG Constraints... it's like Apple took revenge on devs who prefer writing custom views instead of using the Interface Builder. Now both methods are horrible. 😛

GiniVision/Classes/ToolTipView.swift
```diff
@@ -0,0 +1,303 @@
...
+        arrowView.layer.mask = shapeLayer
+        return arrowView
+    }
+}
+
+// MARK: Show and hide tip methods
+
+extension ToolTipView {
+    
+    func show(animations:(() -> ())? = nil){
```
a-szotyori: The purpose of the `animations` block is not obvious. We should either rename it and add a parameter name to be invoked like this for ex. 
```
self.toolTipView?.show(withAuxiliaryAnimations: { // auxiliary just because it sounds fancier than additional :P
    blurEffect.alpha = 1
})
```

We could also extract the animation:
```
UIView.animate(withDuration: 0.5) {
    self.toolTipView?.show()
    blurEffect.alpha = 1
}
```

GiniVision/Classes/ToolTipView.swift
kikettas: I did a mix of both, since Xcode complains because `padding` property has the same as the other one. (I have preferred to not combining instead of using other different name like _paddingSides_ or similar)

GiniVision/Classes/ToolTipView.swift
kikettas: It should only crash to us when we are developing since this class is not public and we control where the tip should appear.
The worst case for the company is that they have the text truncated if they embed the cameraViewController inside a small view (quite small).

GiniVision/Classes/ToolTipView.swift
kikettas: We can always use frameworks like _SnapKit_ that make it more readable, but IMO is not worth.

GiniVision/Classes/ToolTipView.swift
kikettas: You are right, it is not clear.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -118,6 +119,16 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> ()
     fileprivate var cameraFocusLarge: UIImage? {
         return UIImageNamedPreferred(named: "cameraFocusLarge")
     }
+    fileprivate let shouldShowFileImportToolTipUserDefaultsKey = "shouldShowFileImportToolTip"
+    fileprivate var shouldShowFileImportToolTip:Bool {
```
nsobadjiev: This class is already overburdened with a lot of code. I'd try to not have additional logic like that. Maybe you should consider moving it somewhere else.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -245,18 +257,27 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> ()
         camera?.stop()
     }
     
+    public override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        
+        if shouldShowFileImportToolTip {
+            showFileImportTip()
```
nsobadjiev: Why in `viewDidAppear `? It tends to get called a lot more than `viewDidLoad`. If it's because you want to view to be present on screen before showing the tooltip, just add a delay. It would even be nice to have an explicit timeout so we can play with it to see what looks best.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -245,18 +257,27 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> ()
...
         super.viewWillTransition(to: size, with: coordinator)
-        
+
         coordinator.animate(alongsideTransition: { [weak self] _ in
             guard let `self` = self else {
                 return 
             }
             (self.previewView.layer as? AVCaptureVideoPreviewLayer)?.connection?.videoOrientation = self.getVideoOrientation()
-            
+            self.toolTipView?.arrangeViews()
```
nsobadjiev: `arrangeViews ` as a name doesn't really give me a lot of information about what happens there. I assume it re-layouts the view with the new size?

GiniVision/Classes/CameraViewController.swift
```diff
@@ -290,6 +311,36 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> ()
         cameraOverlay.alpha = 0
     }
     
+    fileprivate func showFileImportTip() {
```
nsobadjiev: I'd extract this + the user defaults stuff above to a new file.

GiniVision/Classes/CameraViewController.swift
```diff
@@ -352,6 +403,7 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> ()
     }
     
     @objc fileprivate func importDocument(_ sender: AnyObject) {
+        toolTipView?.dismiss(withCompletion: nil)
```
nsobadjiev: Not really a fan...
What if there's another action method like this? I will have to also remember that I might have to dismiss the tooltip.
I'm not sure if you can find another way, though...

GiniVision/Classes/GiniConfiguration.swift
```diff
@@ -184,6 +184,30 @@ import UIKit
...
+     */
+    public var fileImportToolTipText = "Du kannst jetzt auch ganz einfach Dateien hochladen."
+    
+    /**
+     Sets the text color of the new file import button tip
+     */
+    public var fileImportToolTipTextColor = UIColor.black
+    
+    /**
+     Sets the font of the new file import button tip
```
nsobadjiev: Do you think users will know what "tip" means?

GiniVision/Classes/ToolTipView.swift
```diff
@@ -0,0 +1,300 @@
...
+    fileprivate var closeButtonWidth:CGFloat = 20
+    fileprivate var closeButtonHeight:CGFloat = 20
+    fileprivate var itemSeparation: CGFloat = 16
+    fileprivate var margin:(top:CGFloat, left:CGFloat, right: CGFloat, bottom: CGFloat) = (20, 20, 20, 20)
+    fileprivate var maxWidthOnIpad:CGFloat = 375
+    fileprivate var padding:(top:CGFloat, left:CGFloat, right: CGFloat, bottom: CGFloat) = (16, 16, 16, 16)
+    
+    fileprivate var textWidth:CGFloat {
+        guard let superview = superview else { return 0 }
+        let maxWidth = UIDevice.current.isIpad ? maxWidthOnIpad : superview.frame.width
```
nsobadjiev: I'd love it if a property like `textWidth` doesn't depend on UIDevice and what device the code is running on. Can't you use the superview's frame? Or not have `textWidth` altogether?

GiniVision/Classes/ToolTipView.swift
```diff
@@ -0,0 +1,300 @@
...
+    
+    fileprivate func addShadow() {
+        self.layer.shadowOffset = CGSize(width: 0, height: 2)
+        self.layer.shadowRadius = 0.8
+        self.layer.shadowOpacity = 0.2
+        self.layer.shadowColor = UIColor.black.cgColor
+    }
+    
+    // MARK: Arrange views
+    fileprivate func arrangeFrame(withSuperView superView:UIView?) {
```
nsobadjiev: This method is a monster... and difficult to read. Lets discuss it.

GiniVision/Classes/CameraViewController.swift
kikettas: I thought about that, I decided first to put it in the ToolTipView class itself but I thought that it should be more generic, so I changed my mind. What do you think/suggest?

GiniVision/Classes/CameraViewController.swift
kikettas: I came up with a solution for this, since we need to blur not the entire preview view but only the non-dark frame.

GiniVision/Classes/CameraViewController.swift
nsobadjiev: In my experience, you will never reuse `ToolTipView ` in another project. Also, if you use it in another place, you might still want to only show it once, hence you will still need to keep something in the user defaults. So, I'd be ok with moving that to the `ToolTipView` class.

Alternatively, you should have a dedicated class. For instance, look at `RatingRequirementsStorage` in the gini pay project. However, for the ToolTipView class that's probably an overkill.

GiniVision/Classes/CameraViewController.swift
nsobadjiev: Any comments?

GiniVision/Classes/CameraViewController.swift
kikettas: I extracted the user defaults stuff to the ToolTipView. I had to move the creation of the tip to other method since it could be the case that the onboarding appears and the tooltip has to be hidden, so it will be shown when the onboarding dismisses.

And the creation of the tip cannot be moved to other view since the blurred view depends on it, also I wouldn't move it to other file because all the views on the CameraViewController are created within the class.

GiniVision/Classes/CameraViewController.swift
kikettas: I thought about that but I didn't come up with a better option. The toolTip should only be dismissed if "X" or import buttons are pressed.

GiniVision/Classes/GiniConfiguration.swift
kikettas: Maybe _hint_ is more clear

GiniVision/Classes/CameraViewController.swift
kikettas: It position/arrange views in the correct position.

GiniVision/Classes/ToolTipView.swift
kikettas: I came up with an idea. Thanks! 👍

GiniVision/Classes/ToolTipView.swift
kikettas: I found a bug with the tip position (an edge case) that makes me simplify this method, so maybe now it is easier to read.

Issue #82, opened 2017-10-09
Add custom document validation property


kikettas:
It should be possible to add custom validations for imported documents, apart from the default ones.

PR #83, opened 2017-10-09
Custom document validations

kikettas:
Apart from the default document validations (file size, file type...), there should be possible to add custom validations.

### How to test
Open the example app, go to the camera screen (either Screen or Component API) and try to import a file with a size higher than 5MB.

### Merging
Automatic.

### Ticket 
Issue #82

### DIFF
--- a/Example/GiniVision/SelectAPIViewController.swift
+++ b/Example/GiniVision/SelectAPIViewController.swift
@@ -93,9 +93,11 @@ class SelectAPIViewController: UIViewController {
         giniConfiguration.fileImportSupportedTypes = .pdf_and_images
         giniConfiguration.navigationBarItemTintColor = UIColor.white
         giniConfiguration.customDocumentValidations = { document in
-            print(document.type)
-            let previewImage = document.previewImage
-            print("Validate document")
+            // As an example of custom document validation, we add a more strict check for file size
+            let maxFileSize = 5 * 1024 * 1024
+            if document.data.count > maxFileSize {
+                throw DocumentValidationError.custom(message: "Diese Datei ist leider größer als 5MB")
+            }
         }
         
         // Make sure the app always behaves the same when run from UITests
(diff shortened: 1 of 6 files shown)


GiniVision/Classes/GiniError.swift
```diff
@@ -69,6 +69,31 @@ public enum DocumentValidationError: GiniVisionError{
     /// PDF length exceeded
     case pdfPageLengthExceeded
     
+    /// Custom validation error
+    case custom(message: String)
+    
+    var message:String {
+        switch self {
+        case .exceededMaxFileSize:
+            print()
```
nsobadjiev: What is this `print` here?

GiniVision/Classes/GiniVisionDocument.swift
```diff
@@ -71,6 +71,11 @@ extension GiniVisionDocument {
...
     }
     
+
+    fileprivate var customDocumentValidations: ((GiniVisionDocument) throws -> ())? {
+        return GiniConfiguration.sharedConfiguration.customDocumentValidations
+    }
+    
     // MARK: File validation
     /**
      Validates a document, checking if it has the correct size and type.
```
a-szotyori: For the `- Throws: ` section (cannot select the line here) it would be enough to say, that `DocumentValidationError` is thrown if the document is not valid. Going into details leads to problems like here - implementation was extended without updating the documentation.

Example/GiniVision/SelectAPIViewController.swift
```diff
@@ -92,6 +92,13 @@ class SelectAPIViewController: UIViewController {
         giniConfiguration.debugModeOn = true
         giniConfiguration.fileImportSupportedTypes = .pdf_and_images
         giniConfiguration.navigationBarItemTintColor = UIColor.white
+        giniConfiguration.customDocumentValidations = { document in
```
a-szotyori: Great idea! What if they want to do an async validation? For ex. doing some complex validation that would block the main thread?

PR #87, opened 2017-10-17
Suggestions no results screen

kikettas:
New no results screen that appears when an image has been analyzed and no results have been obtained.

### How to test
Since it is not integrated with both APIs, you cannot see in it in the normal flow but you can show instantiate it and present it whenever you prefer.

### Merging
Automatic.

### Ticket 
Issues #68.


kikettas:
New things should be changed, so it will be closed until everything has been finished.

PR #91, opened 2017-10-20
New onboarding images, texts and an additional page for iPad

kikettas:
Added new onboarding page and updated old images and texts.

### How to test
From the camera screen press "Hilfe" button and see that the new images and texts appear. Also on iPad see that a new page is added.

### Merging
Automatic.

### Ticket 
Issues #90 .


GiniVision/Classes/OnboardingPage.swift
```diff
@@ -18,6 +18,7 @@ import UIKit
     fileprivate var contentView = UIView()
     fileprivate var imageView = UIImageView()
     fileprivate var textLabel = UILabel()
+    fileprivate var needsToRotateImageInLandscape: Bool = false
```
nsobadjiev: Don't really understand that? Why would it NOT rotate?

GiniVision/Classes/OnboardingPage.swift
```diff
@@ -71,6 +73,15 @@ import UIKit
         fatalError("init(coder:) has not been implemented")
     }
     
+    public override func layoutSubviews() {
+        super.layoutSubviews()
+        let rotationAngle:CGFloat = frame.width > frame.height ? -.pi / 2 : 0.0
```
nsobadjiev: Why is this not in the if statement?

GiniVision/Classes/OnboardingPage.swift
kikettas: The imageView is not rotated but the image itself is.

Issue #89, opened 2017-10-19
Supported formats screen


kikettas:
Screen that appears when you choose the option on the help menu.

PR #92, opened 2017-10-20
New supported formats screen

kikettas:
Added new supported formats screen for the help menu.

### How to test
Present the SupportedFormatsViewController.

### Merging
Automatic.

### Ticket 
Issue #89.


Example/Tests/GINISupportedFormatsViewControllerTests.swift
```diff
@@ -0,0 +1,96 @@
...
+    
+    let supportedFormatsViewController = SupportedFormatsViewController(style: .plain)
+    
+    override func setUp() {
+        super.setUp()
+        _ = supportedFormatsViewController.view
+    }
+    
+    func testSectionsCount() {
+        let sectionsCount = supportedFormatsViewController.sections.count
```
nsobadjiev: I don't think this test is enough. It should probably be here as well, but not as the only one.
It just tests that the `numberOfSections` returns `sections.count`. I'd like to see something more concrete... like "Does it contain the 5 sections I expect". I think it's fine to have hardcoded values in the tests. You might want to avoid them if you can, but here, I'd prefer to explicitly test against a concrete expectation about the table.

If you add another section, let the test fail. It would make it explicit that you're changing the class' behaviour.

Example/Tests/GINISupportedFormatsViewControllerTests.swift
```diff
@@ -0,0 +1,96 @@
...
+        let cell = supportedFormatsViewController.tableView.cellForRow(at: indexPath) as! SupportedFormatsTableViewCell
+        let cellImageBackgroundColor = cell.imageBackgroundView.backgroundColor
+        
+        XCTAssertEqual(sectionImageItemBackgroundColor, cellImageBackgroundColor, "cell iage background color should be the same as the one declared on initialization")
+    }
+    
+    func testTableViewCellType() {
+        let indexPath = IndexPath(row: 0, section: 0)
+
+        let cell = supportedFormatsViewController.tableView.cellForRow(at: indexPath)
```
nsobadjiev: You are always testing the cell at (0, 0). You should consider writing a getter method for it so the code is not duplicated in all tests.

Additionally, it might be helpful to also test different cells. Even if they are the same, checking different ones will ensure the same style is applied to all of them.

Example/Tests/GINISupportedFormatsViewControllerTests.swift
```diff
@@ -0,0 +1,96 @@
...
+    
+    func testSectionImageItemBackgroundColor() {
+        let indexPath = IndexPath(row: 0, section: 0)
+
+        let sectionImageItemBackgroundColor = supportedFormatsViewController.sections[indexPath.section].itemsImageBackgroundColor
+        
+        let cell = supportedFormatsViewController.tableView.cellForRow(at: indexPath) as! SupportedFormatsTableViewCell
+        let cellImageBackgroundColor = cell.imageBackgroundView.backgroundColor
+        
+        XCTAssertEqual(sectionImageItemBackgroundColor, cellImageBackgroundColor, "cell iage background color should be the same as the one declared on initialization")
```
nsobadjiev: Cell iage?

GiniVision/Classes/GiniConfiguration.swift
```diff
@@ -467,6 +467,18 @@ import UIKit
...
     
+    // MARK: Supported formats
+    
+    /**
+     Sets the color of the supported formats icon background to the specified color.
+     */
+    public var supportedFormatsIconColor = Colors.Gini.paleGreen
+
+    /**
+     Sets the color of the non supported formats icon background to the specified color.
```
nsobadjiev: I'd say "unsupported"

GiniVision/Classes/SupportedFormatsViewController.swift
```diff
@@ -0,0 +1,116 @@
...
+//
+//  Created by Enrique del Pozo Gómez on 10/19/17.
+//  Copyright © 2017 Gini GmbH. All rights reserved.
+//
+
+import UIKit
+
+final class SupportedFormatsViewController: UITableViewController {
+    
+    let reuseIdentifier = "reuseIdentifier"
```
nsobadjiev: "reuseIdentifier"? That's a bit ambiguous. Why not something like "SupportedFormatsCellIdentifier"?

GiniVision/Classes/SupportedFormatsViewController.swift
```diff
@@ -0,0 +1,116 @@
...
+    let rowHeight: CGFloat = 70
+    let sectionHeight: CGFloat = 70
+    var sections: [(title: String, items: [String], itemsImage:UIImage?, itemsImageBackgroundColor: UIColor)] = [
+        ("Folgende Formate werden unterstützt:",
```
nsobadjiev: Why are these not localized?

GiniVision/Classes/SupportedFormatsViewController.swift
```diff
@@ -0,0 +1,116 @@
...
+//
+
+import UIKit
+
+final class SupportedFormatsViewController: UITableViewController {
+    
+    let reuseIdentifier = "reuseIdentifier"
+    let rowHeight: CGFloat = 70
+    let sectionHeight: CGFloat = 70
+    var sections: [(title: String, items: [String], itemsImage:UIImage?, itemsImageBackgroundColor: UIColor)] = [
```
nsobadjiev: I think this is getting a little too complicated for tuples. Have you considered making a small struct for it?

Also, the `sections` is configurable by the library's users, right? Meaning, they can change it?

GiniVision/Classes/SupportedFormatsViewController.swift
```diff
@@ -0,0 +1,116 @@
...
+        ("Was nicht analysiert wird:",
+         ["Handschrift",
+          "Fotos von Bildschirme"],
+         UIImage(named: "nonSupportedFormatsIcon", in: Bundle(for: GiniVision.self), compatibleWith: nil),
+         GiniConfiguration.sharedConfiguration.nonSupportedFormatsIconColor)
+    ]
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        title = "Unterstützte Formate"
```
nsobadjiev: Localization?

GiniVision/Classes/SupportedFormatsViewController.swift
```diff
@@ -0,0 +1,116 @@
...
+        tableView.register(SupportedFormatsTableViewCell.self, forCellReuseIdentifier: reuseIdentifier)
+        tableView.rowHeight = rowHeight
+        tableView.tableFooterView = UIView()
+        tableView.sectionHeaderHeight = sectionHeight
+        tableView.allowsSelection = false
+        tableView.backgroundColor = Colors.Gini.pearl
+        tableView.alwaysBounceVertical = false
+        
+        if #available(iOS 11.0, *) { // On iOS is .automatic by default.
```
nsobadjiev: Why only on iOS11? Maybe a comment would be helpful?

GiniVision/Classes/SupportedFormatsViewController.swift
```diff
@@ -0,0 +1,116 @@
...
+    
+    override func tableView(_ tableView: UITableView, willDisplayHeaderView view: UIView, forSection section: Int) {
+        if let header = view as? UITableViewHeaderFooterView {
+            header.backgroundView?.backgroundColor = nil
+        }
+    }
+
+}
+
+final class SupportedFormatsTableViewCell: UITableViewCell {
```
nsobadjiev: I think `SupportedFormatsTableViewCell ` is getting too big for being in the same file. If it was just a few lines of code, leaving it in the view controller would have been fine. But in this particular case, I'd put it in a new file.

GiniVision/Classes/SupportedFormatsViewController.swift
kikettas: Yes, I forgot to change it after Xcode created it. 👍

GiniVision/Classes/SupportedFormatsViewController.swift
kikettas: No, the sections are not customizable bu the clients, only the color. 
I think that a `typealias` would be enough for this case.

GiniVision/Classes/SupportedFormatsViewController.swift
kikettas: Because on iOS `contentInsetAdjustmentBehavior` is `automatic` by default and the transition to the target view controller looks weird

GiniVision/Classes/SupportedFormatsViewController.swift
kikettas: Yes, but I think we should start creating folders inside GiniVision project to keep it oganized.

Example/Tests/GINISupportedFormatsViewControllerTests.swift
kikettas: I added some mocked sections so I can test them too, not only the ones embedded in the view controller.

I don't think that is really necessary to create a method just for that, I would leave as it is.

nsobadjiev: Still not convinced. Now the only thing you test is `SupportedFormatsViewController `'s ability to create cells based on the `sections` array.

But you also need to test that it creates the cells specifically needed for the app at runtime. If the specs say "We need three cells with the following contents", you should also test that. What if one cell goes missing? What if the sections array is not set properly at runtime and no cells are shown?

Example/Tests/GINISupportedFormatsViewControllerTests.swift
```diff
@@ -27,22 +27,14 @@ final class GINISupportedFormatsViewControllerTests: XCTestCase {
         _ = supportedFormatsViewController.view
     }
     
-    func testSectionsCount() {
+    func testSectionsMockedCount() {
```
nsobadjiev: You aren ot testing the mocked count, but the count in general. I'd remove the "Mocked" part.

Issue #93, opened 2017-10-20
New open with tutorial screen


kikettas:
Now on help menu you can open a tutorial for open with.

PR #94, opened 2017-10-24
Open with tutorial screen

kikettas:
Added new open with tutorial screen for the help menu.

### How to test
Present the OpenWithTutorialViewController.

### Merging
Automatic.

### Ticket 
Issue #93.


GiniVision/Classes/OpenWithTutorialCollectionHeader.swift
```diff
@@ -0,0 +1,71 @@
...
+//
+//  Created by Enrique del Pozo Gómez on 10/24/17.
+//  Copyright © 2017 Gini GmbH. All rights reserved.
+//
+
+import UIKit
+
+final class OpenWithTutorialCollectionHeader: UICollectionReusableView {
+    
+    let padding:(top: CGFloat, left: CGFloat, bottom: CGFloat, right: CGFloat) = (20, 20, 20, 20)
```
nsobadjiev: Again, this padding thing...

GiniVision/Classes/OpenWithTutorialCollectionHeader.swift
```diff
@@ -0,0 +1,71 @@
...
+        view.backgroundColor = .white
+        return view
+    }()
+    
+    lazy var headerTitle: UILabel = {
+        let label = UILabel()
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.numberOfLines = 0
+        label.textColor = .black
+        let textSize: CGFloat = UIDevice.current.isIpad ? 16 : 14
```
nsobadjiev: This code is everywhere. Can we standardise these? Say that all iPad fonts are 3-4-5 points bigger?

GiniVision/Classes/OpenWithTutorialViewController.swift
```diff
@@ -0,0 +1,117 @@
...
+            self.collectionView?.collectionViewLayout.invalidateLayout()
+        })
+    }
+    
+    private func estimatedCellSize(widthParentSize size: CGSize) -> CGSize {
+        if size.width > size.height && UIDevice.current.isIpad {
+            let width:CGFloat = round(UIScreen.main.bounds.width / CGFloat(self.items.count) - CGFloat(self.stepsCollectionLayout.minimumInteritemSpacing * CGFloat(self.items.count - 1)))
+            return CGSize(width: width, height: UIScreen.main.bounds.height)
+        } else {
+            return CGSize(width: UIScreen.main.bounds.width, height: 550)
```
nsobadjiev: Why not use the view controller view's frame? You'd lose an external dependency.
Also, it would be more correct. When I see something like this, I suspect that there was a problem with changing the view controller size.

Example/Tests/GINIOpenWithTutorialViewControllerTests.swift
```diff
@@ -0,0 +1,113 @@
...
+        
+        XCTAssertEqual(cell.stepIndicator.text, String(describing: indexPath.row + 1), "step indicator for sixth mocked step should be the same as the one declared on initialiation")
+        XCTAssertEqual(cell.stepTitle.text, item.title, "step title for sixth mocked step should be the same as the one declared on initialiation")
+        XCTAssertEqual(cell.stepSubTitle.text, item.subtitle, "step subtitle for sixth mocked step should be the same as the one declared on initialiation")
+        XCTAssertEqual(cell.stepImage.image, item.image, "step image for sixth mocked step should be the same as the one declared on initialiation")
+    }
+
+    func testHeaderDissapearOnLandscape() {
+        let collectionView = openWithTutorialViewController.collectionView!
+        openWithTutorialViewController.view.frame.size = CGSize(width: 1, height: 0) // Simulate landscape
```
nsobadjiev: Manipulating the frame size is not something I'd recommend for orientation changes. It's just a side effect.
How about using the view controller orientation change stuff? Is it possible?

Example/Tests/GINIOpenWithTutorialViewControllerTests.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+    
+    func testFirstStepProperties() {
+        let indexPath = IndexPath(row: 0, section: 0)
+        let item = openWithTutorialViewController.items[indexPath.row]
+        
+        let cell = openWithTutorialViewController.collectionView(openWithTutorialViewController.collectionView!, cellForItemAt: indexPath) as! OpenWithTutorialCollectionCell
+        
+        XCTAssertEqual(cell.stepIndicator.text, String(describing: indexPath.row + 1), "step indicator for first step should be the same as the one declared on initialiation")
+        XCTAssertEqual(cell.stepTitle.text, item.title, "step title for first step should be the same as the one declared on initialiation")
```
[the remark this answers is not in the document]
kikettas: For the next test descriptions I will try to keep it simple.

Example/Tests/GINIOpenWithTutorialViewControllerTests.swift
```diff
@@ -0,0 +1,113 @@
...
+        
+        let cell = openWithTutorialViewController.collectionView(openWithTutorialViewController.collectionView!, cellForItemAt: indexPath) as! OpenWithTutorialCollectionCell
+        
+        XCTAssertEqual(cell.stepIndicator.text, String(describing: indexPath.row + 1), "step indicator for second step should be the same as the one declared on initialiation")
+        XCTAssertEqual(cell.stepTitle.text, item.title, "step title for second step should be the same as the one declared on initialiation")
+        XCTAssertEqual(cell.stepSubTitle.text, item.subtitle, "step subtitle for second step should be the same as the one declared on initialiation")
+        XCTAssertEqual(cell.stepImage.image, item.image, "step image for second step should be the same as the one declared on initialiation")
+    }
+    
+    func testThirdStepProperties() {
```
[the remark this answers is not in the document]
kikettas: That is what I thought at the beginnig but yo made me think... "if for some reason starting on 2 items there is an if that change the title or something in the future?". That is why I agreed with you.
What do you think?

Example/Tests/GINIOpenWithTutorialViewControllerTests.swift
kikettas: I don't like this approach but first I tried to force device/simulator rotation (didn't work), and what I think you suggest (`viewWillTransitionToSize...`) doesn't rotate the interface.
Do you have any better idea? Or do you think we should remove this test since It could be more a UITest?

GiniVision/Classes/Extensions.swift
```diff
@@ -268,4 +268,22 @@ internal extension UIApplication {
     }
 }
 
+internal extension UILabel {
```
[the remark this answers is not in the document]
kikettas: I thought about it, because I prefer to have a different file for each class extension (all inside _Extensions_ folder). I created the issue so I won't forget about it.

GiniVision/Classes/OpenWithTutorialCollectionCell.swift
```diff
@@ -0,0 +1,152 @@
...
+//
+//  Created by Enrique del Pozo Gómez on 10/23/17.
+//  Copyright © 2017 Gini GmbH. All rights reserved.
+//
+
+import UIKit
+
+final class OpenWithTutorialCollectionCell: UICollectionViewCell {
+    
+    let padding:(top: CGFloat, left: CGFloat, bottom: CGFloat, right: CGFloat) = (40, 20, 40, 20)
```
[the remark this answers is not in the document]
kikettas: It refers to the cell itself, since if you call the cell in some place and you want to know the paddign it would be `cell.padding` and not `cell.cellPadding`.

GiniVision/Classes/OpenWithTutorialCollectionHeader.swift
kikettas: Answered above.

GiniVision/Classes/OpenWithTutorialCollectionHeader.swift
kikettas: I had some properties for each kind of text (title, subtitle) and It had a version for iPad and iPhone. 
But I realized that it wasn't useful at all, since in some places the title was 16, in others 14... I wouldn't like to have a `dimen.xml` file now, or at least for this PR (It would be a different PR since this issue is spread in the whole library).

GiniVision/Classes/OpenWithTutorialViewController.swift
kikettas: (https://media.giphy.com/media/U1XhGr8CWqvVC/giphy.gif). I probably moved that from other method and I forgot to change it. 👍

GiniVision/Classes/OpenWithTutorialCollectionCell.swift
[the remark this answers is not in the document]
nsobadjiev: I'm not sure the cell should set its own padding. But ok...

nsobadjiev:
@kikettas there are conflicts - you need to resolve them before we merge.

PR #96, opened 2017-10-25
Suggestions no results screen

kikettas:
Added new no results screen for image analyzed.

### How to test
Present the ImageAnalysisNoResultsViewController.

### Merging
Automatic.

### Ticket 
Issue #68.


Example/Tests/GINIImageAnalysisNoResultsViewControllerTests.swift
```diff
@@ -0,0 +1,140 @@
...
+        _ = viewController.view
+    }
+    
+    func testViewControllerTitleOnInitialization() {
+        let vcTitle = viewController.title
+        
+        XCTAssertEqual(viewControllerTitle, vcTitle, "view controller title should be equals to the one passed in the initialization")
+    }
+    
+    func testSuggestionCollectionMockedItemsCount(){
```
nsobadjiev: You are not testing the **mocked** items. I'd suggest `testSuggestionCollectionItemsCount` as a name

Example/Tests/GINIImageAnalysisNoResultsViewControllerTests.swift
```diff
@@ -0,0 +1,140 @@
...
+    ]
+    
+    override func setUp() {
+        _ = viewController.view
+    }
+    
+    func testViewControllerTitleOnInitialization() {
+        let vcTitle = viewController.title
+        
+        XCTAssertEqual(viewControllerTitle, vcTitle, "view controller title should be equals to the one passed in the initialization")
```
nsobadjiev: Same as the other pull request - the "equals to the one passed in the initialization" is not really important.

Example/Tests/GINIImageAnalysisNoResultsViewControllerTests.swift
```diff
@@ -0,0 +1,140 @@
...
+        let suggestion2Image = viewController.captureSuggestions[indexPath.row].image
+        
+        let suggestionCollectionCell = viewController.collectionView(viewController.suggestionsCollectionView, cellForItemAt: indexPath) as! CaptureSuggestionsCollectionCell
+        
+        XCTAssertEqual(suggestion2Text, suggestionCollectionCell.suggestionText.text, "second suggestionsCollectionView item text should be equal to second captureSuggestions item text")
+        XCTAssertEqual(suggestion2Image, suggestionCollectionCell.suggestionImage.image, "second suggestionsCollectionView item image should be equal to second captureSuggestions item image")
+        
+    }
+    
+    func testThirdSuggestionCell() {
```
nsobadjiev: Same as the other pull request - you don't have to test each cell since the code being run is the same.

Or alternatively, test against the real production data that's going to be shown at runtime.

Example/Tests/GINIImageAnalysisNoResultsViewControllerTests.swift
kikettas: I will try to keep it simple for future PRs.

Example/Tests/GINIImageAnalysisNoResultsViewControllerTests.swift
kikettas: I answered you in the other PR with the reason.

Example/Tests/GINIImageAnalysisNoResultsViewControllerTests.swift
kikettas: But then when I test against the real production data how can I a call the method `testSuggestionCollectionItemsCount`? 😛

PR #101, opened 2017-10-26
Add no results screen to apis

kikettas:
Added new no results screen to Screen API and Component API.

### How to test
 Take a picture that won't have results and see the new screen.

### Merging
Automatic.

### Ticket 
Issues #69 and #70.


Example/GiniVision/SelectAPIViewController.swift
```diff
@@ -22,16 +22,25 @@ class SelectAPIViewController: UIViewController {
         return ProcessInfo.processInfo.arguments.contains("--UITest")
     }
     
-    var analysisDelegate: AnalysisDelegate?
+    var analysisDelegate: AnalysisDelegate? {
```
nsobadjiev: What? Why would setting the delegate result in presenting a screen? It's a side effect.

Example/GiniVision/SelectAPIViewController.swift
```diff
@@ -178,75 +166,58 @@ class SelectAPIViewController: UIViewController {
...
-            present(result, fromDocument: document)
-        }
-    }
-    
     func present(_ result: GINIResult, fromDocument document: GINIDocument) {
         let payFive = ["paymentReference", "iban", "bic", "paymentReference", "amountToPay"]
         let hasPayFive = result.filter { payFive.contains($0.0) }.count > 0
         
         let storyboard = UIStoryboard(name: "Main", bundle: nil)
         if hasPayFive {
```
nsobadjiev: I'd separate the two cases in different methods, this one is getting really big.

Also, the `hasPayFive` name makes me thing all of them need to be there.

Also, Pay Five is something we use internally, but I don't expect others to know what it means. Why not just `hasExtractions`?

GiniVision/Classes/AnalysisContainerViewController.swift
```diff
@@ -64,7 +66,11 @@ internal class AnalysisContainerViewController: UIViewController, ContainerViewC
         let delegate = (navigationController as? GiniNavigationViewController)?.giniDelegate
         delegate?.didCancelAnalysis?()
         
-        _ = navigationController?.popViewController(animated: true)
+        if self == navigationController?.topViewController {
```
nsobadjiev: Another place where you manipulate the navigation stack in a fragile way. Didn't we get rid of those?
And I really don't understand what's happening here.

GiniVision/Classes/AnalysisContainerViewController.swift
```diff
@@ -98,9 +104,29 @@ extension AnalysisContainerViewController: AnalysisDelegate {
...
+            completion(false)
+            return
+        }
+        
+        if document.type == .image {
+            let isCameraViewControllerLoaded = giniNavController.viewControllers.contains(where: { viewController in
+                return viewController is CameraContainerViewController
+            })
+        
+            giniNavController.pushViewController(ImageAnalysisNoResultsContainerViewController(canGoBack: isCameraViewControllerLoaded), animated: true)
```
nsobadjiev: Another one... and I really dislike that `canGoBack `.
**It's not the view controller's responsibility to know if it can go back or not.**

GiniVision/Classes/AnalysisViewController.swift
```diff
@@ -23,6 +23,15 @@ import UIKit
...
     func displayError(withMessage message: String?, andAction action: NoticeAction?)
+    
+    /**
+     In case that the `GiniVisionDocument` analysed is an image it will display a no results screen
+     with some capture suggestions. It won't show any screen if it is not an image, pointing out
+     on the completion handler that any screen was shown.
+     
+     - parameter completion: Handler that returns `true` if a screen was shown or `false` if it wasn't.
+     */
+    func displayNoResultsScreen(completion: ((Bool) -> ()))
```
nsobadjiev: I really dislike this. I method called `displayNoResultsScreen` should, in fact, display the screen.

Right now, it says "If you tell me to display something, I'll give you back a result saying if I decided to listen to you, or not". It's an OOP principles to **tell** objects what to do, not ask them.

GiniVision/Classes/AnalysisContainerViewController.swift
```diff
@@ -98,9 +104,29 @@ extension AnalysisContainerViewController: AnalysisDelegate {
     
     func displayError(withMessage message: String?, andAction action: NoticeAction?) {
         let notice = NoticeView(text: message ?? "", noticeType: .error, action: action)
-        DispatchQueue.main.async { 
+        DispatchQueue.main.async {
             self.showNotice(notice)
         }
     }
     
+    func displayNoResultsScreen(completion: ((_ shown: Bool) -> ())) {
```
nsobadjiev: Why are you using a completion closure when you always call it synchronously before the method returns? Just return a boolean.

GiniVision/Classes/GiniVision.swift
```diff
@@ -17,30 +17,40 @@ import UIKit
 @objc public protocol GiniVisionDelegate {
     
     /**
+     Called when the user has taken a picture or imported a file (image or PDF) from camera roll or document explorer
```
nsobadjiev: If the user imported a file... it's not capturing anymore, is it? I guess, the name `didCapture` is misleading.

GiniVision/Classes/GiniVision.swift
```diff
@@ -61,6 +71,10 @@ import UIKit
     
 }
 
+extension GiniVisionDelegate {
```
nsobadjiev: An empty extension?

GiniVision/Classes/GiniVision.swift
```diff
@@ -106,7 +120,14 @@ import UIKit
...
                 viewController = AnalysisContainerViewController(document: document)
             }
-            delegate.didImport?(document)
+            
+            if let didCapture = delegate.didCapture(document:) {
+                didCapture(document)
+            } else if let didCapture = delegate.didCapture(_:) {
+                didCapture(document.data)
+            } else {
+                fatalError("GiniVisionDelegate.didCapture(document: GiniVisionDocument) should be implemented")
```
nsobadjiev: `didCapture(document: GiniVisionDocument)` or `didCapture(_:)`?

GiniVision/Classes/ImageAnalysisNoResultsContainerViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+            setupLeftNavigationItem(usingResources: closeButtonResources, selector: #selector(close))
+        }
+        
+        // Configure view hierachy
+        view.addSubview(containerView)
+        
+        // Add constraints
+        addConstraints()
+        
+        // remove analysis from stack.
```
nsobadjiev: Did you forget to remove that?

GiniVision/Classes/ImageAnalysisNoResultsContainerViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+        if canGoBack {
+            setupLeftNavigationItem(usingResources: backButtonResources, selector: #selector(backToCamera))
+        } else {
+            setupLeftNavigationItem(usingResources: closeButtonResources, selector: #selector(close))
+        }
+        
+        // Configure view hierachy
+        view.addSubview(containerView)
+        
+        // Add constraints
```
nsobadjiev: A useless comment.

GiniVision/Classes/ImageAnalysisNoResultsContainerViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+        view.backgroundColor = GiniConfiguration.sharedConfiguration.backgroundColor
+        
+        // Configure close button
+        if canGoBack {
+            setupLeftNavigationItem(usingResources: backButtonResources, selector: #selector(backToCamera))
+        } else {
+            setupLeftNavigationItem(usingResources: closeButtonResources, selector: #selector(close))
+        }
+        
+        // Configure view hierachy
```
nsobadjiev: Useless and also not true. You are not configuring the hierarchy. You are just adding a subview.

GiniVision/Classes/ImageAnalysisNoResultsContainerViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+        
+        imageAnalysisNoResultsViewController.didTapBottomButton = { [weak self] in
+            self?.backToCamera()
+        }
+        contentController = imageAnalysisNoResultsViewController
+        
+        // Configure colors
+        view.backgroundColor = GiniConfiguration.sharedConfiguration.backgroundColor
+        
+        // Configure close button
```
nsobadjiev: Too vague. Either don't have anything at all, or explain why there are two options.

GiniVision/Classes/ImageAnalysisNoResultsContainerViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        // Add content to container view
```
nsobadjiev: Useless

GiniVision/Classes/ImageAnalysisNoResultsContainerViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+        displayContent(contentController) 
+    }
+    
+    @objc private func backToCamera() {
+        let delegate = (navigationController as? GiniNavigationViewController)?.giniDelegate
```
nsobadjiev: Again, the view controller should be so aware of its "surroundings". It should know that it's in a navigation stack of a `GiniNavigationViewController `, that happens to have a delegate. Please use another delegate (or something else) to relay information between these classes.

GiniVision/Classes/ImageAnalysisNoResultsContainerViewController.swift
```diff
@@ -0,0 +1,94 @@
...
+        delegate?.didCancelAnalysis?()
+        
+        if let cameraViewController = navigationController?.viewControllers.flatMap({ $0 as? CameraContainerViewController }).first {
```
nsobadjiev: Yet another place where you manipulate the navigation stack ad-hoc.

GiniVision/Classes/ReviewContainerViewController.swift
```diff
@@ -129,11 +129,18 @@ internal class ReviewContainerViewController: UIViewController, ContainerViewCon
     }
     
     @IBAction func analyse() {
-        let delegate = (self.navigationController as? GiniNavigationViewController)?.giniDelegate
-        delegate?.didReview(document!.data, withChanges: changes)
+        guard let delegate = (self.navigationController as? GiniNavigationViewController)?.giniDelegate, let document = document else { return }
```
nsobadjiev: Annnd, another one.

GiniVision/Classes/AnalysisContainerViewController.swift
kikettas: We didn't for the Screen API, since we still use the containers. You cannot pop a view controller inside a navigation controller if it is the rootViewController, that is way I need to check if the last one.
But this is wrong, it should be the rootViewController not the topViewController.

GiniVision/Classes/AnalysisContainerViewController.swift
kikettas: Because clients could still use the current navigation controller, and this way they can concatenate the same animation. I really don't like this, but it is the best way (IMO) for now.

GiniVision/Classes/AnalysisContainerViewController.swift
kikettas: It is the only way using the Screen API, using the containers...

GiniVision/Classes/AnalysisViewController.swift
kikettas: Yes, probably the name is wrong, but I think it is the only way to proceed. Do you suggest a better name?

GiniVision/Classes/GiniVision.swift
kikettas: I thought about that, but I wasn't sure that `didPick` was a better option. What do you think?

GiniVision/Classes/GiniVision.swift
kikettas: didCapture(document: GiniVisionDocument) since it is the new one. I tried first to call it `didCapture(_ document: GiniVisionDocument)` but Xcode saw it as an ambiguity.

GiniVision/Classes/ImageAnalysisNoResultsContainerViewController.swift
kikettas: Not, I copied this method from other Container and I think I forgot to remove all these stupid comments.

Example/GiniVision/SelectAPIViewController.swift
kikettas: It is getting really big yes 👍

Example/GiniVision/SelectAPIViewController.swift
kikettas: It made sense before changing the didReview method, but now it is only changed in one place so it doesn't make sense anymore. 👍

GiniVision/Classes/ImageAnalysisNoResultsContainerViewController.swift
kikettas: I know it should not, but as I said above, it is the only way for now.

GiniVision/Classes/ImageAnalysisNoResultsContainerViewController.swift
kikettas: Answered above.

GiniVision/Classes/ReviewContainerViewController.swift
kikettas: Answered above.

GiniVision/Classes/AnalysisContainerViewController.swift
nsobadjiev: concatenate the same animation? I don't understand.

kikettas: I was wrong. From that point they can still use the same UINavigationController, so I can just a return a boolean.

GiniVision/Classes/AnalysisViewController.swift
nsobadjiev: No name will fix this issue. But maybe the name should reflect that the call might fail. Like `tryDisplayNoResultsScreen`.

GiniVision/Classes/GiniVision.swift
nsobadjiev: `didImportImage`?
`didChooseImage`?

GiniVision/Classes/GiniVision.swift
nsobadjiev: I meant that you should mention both methods in the hint message. Otherwise, it's not true. The assertion will fail only if both are not implemented.

GiniVision/Classes/ImageAnalysisNoResultsContainerViewController.swift
nsobadjiev: 👎

GiniVision/Classes/GiniVision.swift
kikettas: This is also called when a image is captured, so they are not correct.

PR #105, opened 2017-10-30
Split extension file into several files.

kikettas:
Split extension file into several files.

### How to test
Nothing to test.

### Merging
Automatic.

### Ticket 
Issue #97.


kikettas:
Apart that is not Swift-like, some extensions has more than one purpose. Having all extensions inside `Extensions` folder make it more clear.

PR #110, opened 2017-11-07
Added jenkinsfile

kikettas:
Added Jenkinsfile to the project.

### How to test
Nothing to test.

### Merging
Automatic.


Documentation/deploy-documentation.sh
```diff
@@ -1,37 +1,22 @@
...
-pip install -r requirements.txt
-
-make clean
-make html
-
-cd build
+cd Documentation/
 rm -rf gh-pages
-git clone -b gh-pages https://${DOC_PUSH_TOKEN}@github.com/gini/gini-vision-lib-ios.git gh-pages
+git clone -b gh-pages https://"$github_user":"$github_password"@github.com/gini/gini-vision-lib-ios.git gh-pages
```
nsobadjiev: Doesn't the Jenkins machine have an SSH key for Github? It might be possible to use SSH here and skip those credentials.

Example/GiniVision.xcodeproj/xcshareddata/xcschemes/GiniVision-Example.xcscheme
```diff
@@ -42,7 +42,7 @@
             buildForAnalyzing = "YES">
             <BuildableReference
                BuildableIdentifier = "primary"
-               BlueprintIdentifier = "FA4C1717ACBA7DAFF610065C12D00D2C"
+               BlueprintIdentifier = "0F25960EE62BCB781A7B936531602322"
```
nsobadjiev: What is this?

Jenkinsfile
```diff
@@ -0,0 +1,57 @@
+pipeline {
+  agent any
+  environment {
+    GIT = credentials('github')
+  }
+  stages {
+    stage('Prerequisites') {
+      steps {
+        sh '/usr/local/bin/pod install --project-directory=Example/'
```
nsobadjiev: Instead of using the full path, can you use an environment? Like here:
https://github.com/gini/gini-switch-sdk-android/blob/develop/Jenkinsfile#L65

Documentation/deploy-documentation.sh
kikettas: I supposed that It doesn't, since it the other repo it was not used. I think it is fine to use both user and password, so we don't have to worry about setting up the ssh key.

Example/GiniVision.xcodeproj/xcshareddata/xcschemes/GiniVision-Example.xcscheme
kikettas: Weird Xcode stuff. I imagine that after trying several signing configurations, it changed.

Jenkinsfile
kikettas: I saw it, but for me setting the absolute path is more clear.

PR #114, opened 2017-11-16
Add settings to example app

kikettas:
Added settings menu in the select api screen, so it is possible to test different configuration without having to run and build the app again.

### How to test
Press in the metadata text in the select api screen and see that a new screen appears. Try to change one of the settings and then see that it is applied both in the Screen API and Component API

### Merging
Automatic.

### Ticket 
Issue #109


Example/GiniVision/SelectAPIViewController.swift
```diff
@@ -28,7 +29,7 @@ enum GiniVisionAPIType {
  */
 final class SelectAPIViewController: UIViewController {
     
-    @IBOutlet weak var metaInformationLabel: UILabel!
+    @IBOutlet weak var metaInformationLabel: UIButton!
```
nsobadjiev: Can you change the name to `metaInformationButton`?

PR #120, opened 2017-12-04
Add QR Scanning in Camera

kikettas:
Added QR scanning in Camera class.

### How to test
Scan a QR code and see that is returned in the captureOutput delegate.

### Merging
Automatic.

### Issues
#122

### DIFF
--- a/Example/Tests/GINICameraViewControllerTests.swift
+++ b/Example/Tests/GINICameraViewControllerTests.swift
@@ -3,9 +3,10 @@ import XCTest
 
 class CameraViewControllerTests: XCTestCase {
     
-    var vc = CameraViewController(successBlock: { _ in }, failureBlock: { _ in })
+    var vc: CameraViewController!
     
     func testInitialization() {
+        vc = CameraViewController(successBlock: { _ in }, failureBlock: { _ in })
         XCTAssertNotNil(vc, "view controller should not be nil")
     }
     
@@ -33,6 +34,27 @@ class CameraViewControllerTests: XCTestCase {
         
         XCTAssertFalse(vc.captureButton.isEnabled, "capture button should be disaled when tooltip is shown")
         
+    }
+    
+    func testCaptureImage() {
+        let expect = expectation(description: "image is captured")
+        vc = CameraViewController(successBlock: { document in
+            XCTAssertTrue(document.type == .image, "document should be an image")
+            expect.fulfill()
+        }, failureBlock: { _ in
+            XCTFail()
+            expect.fulfill()
+        })
+        
+        _ = vc.view
+        vc.viewWillAppear(true)
+        vc.captureButton.sendActions(for: .touchUpInside)
+        
+        waitForExpectations(timeout: 5.0) { error in
+            if let error = error {
+                XCTFail("An error ocurried when capturing image: \(error)")
+            }
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -34,5 +35,26 @@ class CameraViewControllerTests: XCTestCase {
...
+        }, failureBlock: { _ in
+            XCTFail()
+            expect.fulfill()
+        })
+        
+        _ = vc.view
+        vc.viewWillAppear(true)
+        vc.captureButton.sendActions(for: .touchUpInside)
+        
+        waitForExpectations(timeout: 5.0) { error in
```
nsobadjiev: Why so long? Should it be enough to schedule it during the next runloop frame? What I'm getting at, is that if it fails, it better fail immediately. Tests should be fast.

GiniVision/Classes/Camera.swift
```diff
@@ -55,89 +77,101 @@ internal class Camera {
...
+extension Camera: AVCaptureMetadataOutputObjectsDelegate {
+    func captureOutput(_ output: AVCaptureOutput!,
+                       didOutputMetadataObjects metadataObjects: [Any]!,
+                       from connection: AVCaptureConnection!) {
+        if metadataObjects.isEmpty {
+            return
+        }
         
-        self.session.commitConfiguration()
+        if let metadataObj = metadataObjects[0] as? AVMetadataMachineReadableCodeObject,
```
nsobadjiev: Instead of checking if the array is empty and then getting the first object, can't you do a `if let metadataObjects.first`?

GiniVision/Classes/Camera.swift
```diff
@@ -55,89 +77,101 @@ internal class Camera {
...
+            metadataObj.type == AVMetadataObjectTypeQRCode {
+            // Create Gini Document
```
nsobadjiev: So why aren't you creating it?

GiniVision/Classes/CameraViewController.swift
```diff
@@ -353,18 +369,16 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> ()
...
-                let imageDocument = GiniImageDocument(data: imageData, imageSource: .camera, deviceOrientation: UIApplication.shared.statusBarOrientation)
-                
-                // Call success block
-                self.successBlock?(imageDocument)
-            } catch let error as CameraError {
-                self.failureBlock?(error)
-            } catch _ {
-                print("GiniVision: An unknown error occured.")
+        camera.captureStillImage {[weak self] imageData, error in
+            guard let imageData = imageData, error == nil else {
```
nsobadjiev: It's a styling thing, but wouldn't it be more readable to have the error check on a separate line?

Example/Tests/GINICameraViewControllerTests.swift
kikettas: Yes, you are right. It should take less than 1sec. 👍

GiniVision/Classes/Camera.swift
kikettas: Because I decided to split the QR feature in two PRs, scanning and processing. This way I don't have two big PRs

GiniVision/Classes/Camera.swift
kikettas: I kept it like this since I will probably change it in the QR processing branch.

GiniVision/Classes/CameraViewController.swift
kikettas: Yes, it would be more readable. 👍

PR #124, opened 2017-12-08
QR Processing

kikettas:
Added QR processing, with a custom view por the popup in the camera screen.

### How to test
Scan a valid QR code and see that a popup appears, being able to proceed to results.

### Merging
Automatic.

### Issues
#121

### DIFF
--- a/GiniVision/Assets/Localizable.strings
+++ b/GiniVision/Assets/Localizable.strings
@@ -27,8 +27,8 @@
 "ginivision.camera.documentValidationError.tooManyPages" = "Das Dokument darf nicht mehr als 10 Seiten haben";
 "ginivision.camera.documentValidationError.wrongFormat" = "Nur Dateien vom Format jpg, tiff, png, gif oder pdf können analysiert werden";
 "ginivision.camera.fileImportTip" = "Es können jetzt auch ganz einfach Dateien hochgeladen werden.";
-"ginivision.camera.qrCodeDetectedPopup.message" = "QR detected!";
-"ginivision.camera.qrCodeDetectedPopup.buttonTitle" = "Use it!";
+"ginivision.camera.qrCodeDetectedPopup.message" = "QR verwenden?";
+"ginivision.camera.qrCodeDetectedPopup.buttonTitle" = "Hier geht’s weiter.";
 
 "ginivision.onboarding.firstPage"  = "Seite vollständig in den angezeigten Rahmen einpassen";
 "ginivision.onboarding.secondPage" = "Seite glatt streichen";
--- a/GiniVision/Assets/Localizable.strings
+++ b/GiniVision/Assets/Localizable.strings
@@ -27,6 +27,8 @@
 "ginivision.camera.documentValidationError.tooManyPages" = "Das Dokument darf nicht mehr als 10 Seiten haben";
 "ginivision.camera.documentValidationError.wrongFormat" = "Nur Dateien vom Format jpg, tiff, png, gif oder pdf können analysiert werden";
 "ginivision.camera.fileImportTip" = "Es können jetzt auch ganz einfach Dateien hochgeladen werden.";
+"ginivision.camera.qrCodeDetectedPopup.message" = "QR detected!";
+"ginivision.camera.qrCodeDetectedPopup.buttonTitle" = "Use it!";
 
 "ginivision.onboarding.firstPage"  = "Seite vollständig in den angezeigten Rahmen einpassen";
 "ginivision.onboarding.secondPage" = "Seite glatt streichen";
--- a/GiniVision/Classes/GiniQRCodeDocument.swift
+++ b/GiniVision/Classes/GiniQRCodeDocument.swift
@@ -11,7 +11,7 @@ import Foundation
     public var type: GiniVisionDocumentType = .qrcode
     public var data: Data
     public lazy var previewImage: UIImage? = {
-       return self.generateQRCodeImage(from: self.data)
+        return self.generateQRCodeImage(from: self.data)
     }()
     public var isReviewable: Bool = false
     public var isImported: Bool = false
@@ -100,8 +100,8 @@ extension GiniQRCodeDocument {
                 parameters["paymentReference"] = paymentReference
             }
             if let amount = queryParameters["amount"] as? String,
-                let currency = queryParameters["currency"] as? String,
-                let amountNormalized = normalize(amount: amount, currency: currency) {
+                let amountNormalized = normalize(amount: amount,
+                                                 currency: queryParameters["currency"] as? String ?? "EUR") {
                 parameters["amountToPay"] = amountNormalized
             }
         }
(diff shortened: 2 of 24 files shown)


GiniVision/Assets/Localizable.strings
```diff
@@ -27,6 +27,8 @@
 "ginivision.camera.documentValidationError.tooManyPages" = "Das Dokument darf nicht mehr als 10 Seiten haben";
 "ginivision.camera.documentValidationError.wrongFormat" = "Nur Dateien vom Format jpg, tiff, png, gif oder pdf können analysiert werden";
 "ginivision.camera.fileImportTip" = "Es können jetzt auch ganz einfach Dateien hochgeladen werden.";
+"ginivision.camera.qrCodeDetectedPopup.message" = "QR detected!";
```
nsobadjiev: Shouldn't that be in german?

GiniVision/Classes/GiniQRCodeDocument.swift
```diff
@@ -0,0 +1,162 @@
...
+        super.init()
+    }
+    
+    public func checkType() throws {
+        if self.qrCodeFormat == nil || self.extractedParameters.isEmpty {
+            throw DocumentValidationError.qrCodeFormatNotValid
+        }
+    }
+    
+    fileprivate func generateQRCodeImage(from data: Data) -> UIImage? {
```
nsobadjiev: Why is there an image generation method in a model class?
Transitively, you have a dependency yo UIKit is a model class (UIImage).

GiniVision/Classes/GiniQRCodeDocument.swift
```diff
@@ -0,0 +1,162 @@
...
+        let filter = CIFilter(name: "CIQRCodeGenerator")
+        
+        filter?.setValue(data, forKey: "inputMessage")
+        filter?.setValue("Q", forKey: "inputCorrectionLevel")
+        
+        if let outputImage = filter?.outputImage {
+            return UIImage(ciImage: outputImage.applying(CGAffineTransform(scaleX: 2, y: 2)))
```
nsobadjiev: Are you sure you can hardcode the scale? Wouldn't it be 3 if it was an iPhone 7 Plus, for instance?

GiniVision/Classes/GiniQRCodeDocument.swift
```diff
@@ -0,0 +1,162 @@
...
+        if let object = object as? GiniQRCodeDocument {
+            return self.scannedString == object.scannedString
+        }
+        return false
+    }
+}
+
+// MARK: Extractions
+
+extension GiniQRCodeDocument {
```
nsobadjiev: Not an extension, really. It's a full blown class that handles QR codes. Seems out of place here.

GiniVision/Classes/GiniQRCodeDocument.swift
```diff
@@ -0,0 +1,162 @@
...
+        
+        if !lines[4].isEmpty {
+            parameters["bic"] = lines[4]
+        }
+        
+        if !lines[5].isEmpty {
+            parameters["paymentRecipient"] = lines[5]
+        }
+        
+        if !lines[9].isEmpty {
```
nsobadjiev: How can you be sure all those 9 lines exist?

GiniVision/Classes/GiniVision.swift
```diff
@@ -31,6 +31,12 @@ import UIKit
     @available(*, deprecated)
     @objc optional func didCapture(_ imageData: Data)
     
+    /**
+     Called when a QR Code has been detect in the camera screen and the user wants to proceed with it.
```
nsobadjiev: detected*

GiniVision/Classes/GiniVisionDocument.swift
```diff
@@ -12,27 +12,31 @@ import Foundation
...
     func checkType() throws
 }
 
 // MARK: GiniVisionDocumentType
 
-@objc public enum GiniVisionDocumentType:Int {
+@objc public enum GiniVisionDocumentType: Int {
     case pdf = 0
     case image = 1
+    case qrcode = 2
```
nsobadjiev: Can't you omit these values (0, 1, 2). It's should be the default.

GiniVision/Classes/GiniVisionDocument.swift
```diff
@@ -12,27 +12,31 @@ import Foundation
...
 }
 
 // MARK: GiniVisionDocumentBuilder
 
 /**
- * The `GiniVisionDocumentBuilder` provides a way to build a `GiniVisionDocument` from a `Data` object. The `DocumentSource` must be provided in the initialization, being optional but highly recommended setting the `DocumentImportMethod` afterwards. This could be an example of how a `GiniVisionDocument` should be built when it has been imported with the _Open with_ feature.
+ The `GiniVisionDocumentBuilder` provides a way to build a `GiniVisionDocument` from a `Data` object.
+ The `DocumentSource` must be provided in the initialization, being optional but highly recommended
```
nsobadjiev: You cannot say "must" and then say it's optional. Use "should" then.

GiniVision/Classes/GiniVisionDocument.swift
```diff
@@ -12,27 +12,31 @@ import Foundation
...
+ setting the `DocumentImportMethod` afterwards. This could be an example of how a `GiniVisionDocument`
+ should be built when it has been imported with the _Open with_ feature.
```
nsobadjiev: I really don't understand what you mean with these several sentences. Can you paraphrase?

GiniVision/Classes/QRCodeDetectedPopupView.swift
```diff
@@ -0,0 +1,170 @@
...
+    }
+    
+    func hide(after seconds: TimeInterval = 0, completion: (() -> Void)? = nil) {
+        DispatchQueue.main.asyncAfter(deadline: .now() + seconds, execute: { [weak self] in
+            guard let `self` = self, let superview = self.superview else { return }
+            self.bottomConstraint?.constant = self.imageSize.height +
+                self.padding.top +
+                self.padding.bottom +
+                self.margin.bottom
+            UIView.animate(withDuration: 0.5,
```
nsobadjiev: This 0.5 is duplicated with the show method. Also, I keep recommending to have animation durations set as parameters (slow, medium, fast).

GiniVision/Assets/Localizable.strings
kikettas: Yes, I was waiting for the german version that is why I left these in english.

GiniVision/Classes/GiniQRCodeDocument.swift
kikettas: I was thinking about this since I wrote it, I wasn't sure about it. Yes, definitely this should me move to other class.

GiniVision/Classes/GiniQRCodeDocument.swift
kikettas: I'm just making it slightly bigger, since it is really small by default. In my opinion, 2x scale is more than enough

GiniVision/Classes/GiniQRCodeDocument.swift
kikettas: Because I checked it when the document is validated.

GiniVision/Classes/GiniQRCodeDocument.swift
kikettas: In this the methods are quite specific por this type of QR processing, since there two types (behzal and epc0869). At the beginning I though to move this to a different class, but this is not handling generic qr codes.

GiniVision/Classes/GiniVisionDocument.swift
kikettas: I hate it but we cannot get rid of them because `objc` enums must declare an integer raw type. And this should be a `objc` enum since it is used in an `objc` protocol (`GiniVisionDocument`), which is needed in an optional `objc` method (in `GiniVisionDelegate`)

GiniVision/Classes/GiniQRCodeDocument.swift
nsobadjiev: Still, it mixes business logic with the modal. I'd still say, move it.

PR #127, opened 2017-12-21
New credentials injection method 

kikettas:
New credentials injection method in Example app.

### How to test
Add the environment variables and check that GVL Example app analyze documents.

### Merging
Automatic.


Jenkinsfile
```diff
@@ -53,6 +53,11 @@ pipeline {
         sh 'rm -rf build'
         sh 'rm Example/Release-keys.xcconfig'
       }
+      post {
+        always {
+          sh 'rm -rf Example/Keys.plist || true'
```
nsobadjiev: You don't need the `-r`, maybe not even the `-f`. It's not a directory so it's not recursive and I doubt you need to force it.

Example/README.md
```diff
@@ -4,4 +4,4 @@ The Swift example app provides you with a sample usage of the Gini Vision Librar
 
 The Gini API SDK for iOS is used for analyzing documents and sending feedback.
 
-Before analyzing documents, you need to set your Gini API Client Id and Secret in the `ENV-Keys.Example.xcconfig` and rename it according to the configurations you want to use.
\ No newline at end of file
+Before analyzing documents, you need to set your Gini API Client ID and Client Password in your Environment Variables (`Xcode -> Product -> Scheme -> Edit scheme -> Environment Variables)` as *client_id* and *client_password*, respectively.
```
nsobadjiev: Why aren't you also explaining how people can use the keys file?

Example/Keys.plist
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nsobadjiev: There's no key here. It's credentials, or client secrets. The file name is misleading.

Example/GiniVision/SelectAPIViewController.swift
```diff
@@ -37,8 +37,7 @@ final class SelectAPIViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         
-        let customClientId = UserDefaults.standard.string(forKey: kSettingsGiniSDKClientIdKey) ?? ""
-        let clientId = customClientId != "" ? customClientId : kGiniClientId
+        let clientId = ProcessInfo.processInfo.environment["client_id"] ?? ""
```
nsobadjiev: What about the file? Why are you only reading from the ENV?

Example/GiniVision/DocumentService.swift
```diff
@@ -82,14 +93,6 @@ final class DocumentService {
         print("Gini Vision Library for iOS (\(GiniVision.versionString)) / Client id: \(clientId)")
     }
     
-    func populateSettingsPage() {
```
nsobadjiev: No settings page anymore?

.gitignore
```diff
@@ -19,6 +19,11 @@ DerivedData
 *.hmap
 *.ipa
 .env
+*.xcodeproj/*
+!*.xcodeproj/project.pbxproj
+!*.xcodeproj/xcshareddata/
+!*.xcworkspace/contents.xcworkspacedata
+/*.gcno
```
nsobadjiev: Don't you need an ignore entry for the keys file?

kikettas: No, it is empty and I think it should be there.

Example/GiniVision/DocumentService.swift
kikettas: No one was using it, it wasn't even documented. And adding a password in the settings is not the best way I think.

.gitignore
nsobadjiev: What if I decide to use it, fill it with valid credentials and commit without realising it contains sensitive data?

Example/GiniVision/SelectAPIViewController.swift
kikettas: Yes, I forgot about this one.

.gitignore
kikettas: Yes, you are right. 👍

PR #140, opened 2018-01-22
Final UI fixes with QR codes.

kikettas:
Changed behavior of the QR code scanning (now a QR code can be scanned again after the first 10 seconds), fixed issue with Component API and added mandatory iban for QR code validation.

### How to test
Run the example app, scan a QR code and wait 10 seconds until it disappears. 

### Merging
Automatic.
 
### Issues
Issue #139

### DIFF
--- a/GiniVision/Classes/CameraViewController.swift
+++ b/GiniVision/Classes/CameraViewController.swift
@@ -121,9 +121,7 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> Void
         return FilePickerManager()
     }()
     fileprivate var detectedQRCodeDocument: GiniQRCodeDocument?
-    fileprivate var currentQRCodePopup: QRCodeDetectedPopupView? {
-        return self.view.subviews.flatMap { $0 as? QRCodeDetectedPopupView }.first
-    }
+    fileprivate var currentQRCodePopup: QRCodeDetectedPopupView?
 
     var didShowCamera: (() -> Void)?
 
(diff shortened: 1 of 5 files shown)


GiniVision/Classes/CameraViewController.swift
```diff
@@ -121,6 +121,10 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> Void
         return FilePickerManager()
     }()
     fileprivate var detectedQRCodeDocument: GiniQRCodeDocument?
+    fileprivate var currentQRCodePopup: QRCodeDetectedPopupView? {
+        return self.view.subviews.flatMap { $0 as? QRCodeDetectedPopupView }.first
```
nsobadjiev: You know how much I hate going through the view hierarchy in order to find something. If you need a view, hold a reference to it as you create it.

nsobadjiev: Also, why would you have several?

kikettas: Yes, you are right. Although it is nothing in terms of performance to go through the view hierarchy with the `flatMap`, I think It is more clear if it is stored in the class (uglier though).

GiniVision/Classes/CameraViewController.swift
```diff
@@ -121,9 +121,7 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> Void
         return FilePickerManager()
     }()
     fileprivate var detectedQRCodeDocument: GiniQRCodeDocument?
-    fileprivate var currentQRCodePopup: QRCodeDetectedPopupView? {
-        return self.view.subviews.flatMap { $0 as? QRCodeDetectedPopupView }.first
-    }
+    fileprivate var currentQRCodePopup: QRCodeDetectedPopupView?
```
nsobadjiev: Maybe it makes sense to have it as weak?

Issue #144, opened 2018-02-14
Status bar style setting


kikettas:
Add status bar style setting to Gini Configuration.

PR #145, opened 2018-02-14
Feature custom status bar

kikettas:
Added status bar style setting to GiniConfiguration, so now it can be customized. It works both when `UIViewControllerBasedStatusBarAppearance` is enabled and disabled. 

### How to test
Run the example app with `UIViewControllerBasedStatusBarAppearance` as NO in the Info.plist file. Check that status bar changes once the GVL is launched.
Do the same but with `UIViewControllerBasedStatusBarAppearance` as `YES` and check that it is also working.

### Merging
Automatic.
 
### Issues
Issue #144

### DIFF
--- a/GiniVision/Classes/GiniConfiguration.swift
+++ b/GiniVision/Classes/GiniConfiguration.swift
@@ -188,6 +188,15 @@ import UIKit
      */
     public var qrCodeScanningEnabled = false
     
+    /**
+     Indicates the status bar style on the Gini Vision Library.
+     
+     - note: If `UIViewControllerBasedStatusBarAppearance` is set to `false` in the `Info.plist`,
+     it may not work in further versions of iOS since `UIApplication.setStatusBarStyle` method was
+     deprecated on iOS 9.0
+     */
+    public var statusBarStyle = UIStatusBarStyle.lightContent
+    
     // MARK: Camera options
     
     /**
--- a/GiniVision/Classes/GiniConfiguration.swift
+++ b/GiniVision/Classes/GiniConfiguration.swift
@@ -189,10 +189,10 @@ import UIKit
     public var qrCodeScanningEnabled = false
     
     /**
-     Indicates the status bar style on the Gini Vision Library.
+     Indicates the status bar style in the Gini Vision Library.
      
      - note: If `UIViewControllerBasedStatusBarAppearance` is set to `false` in the `Info.plist`,
-     it may not work in further versions of iOS since `UIApplication.setStatusBarStyle` method was
+     it may not work in future versions of iOS since the `UIApplication.setStatusBarStyle` method was
      deprecated on iOS 9.0
      */
     public var statusBarStyle = UIStatusBarStyle.lightContent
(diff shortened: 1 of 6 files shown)


GiniVision/Classes/GiniConfiguration.swift
```diff
@@ -188,6 +188,15 @@ import UIKit
      */
     public var qrCodeScanningEnabled = false
     
+    /**
+     Indicates the status bar style on the Gini Vision Library.
```
nsobadjiev: **in** the Gini Vision Library

GiniVision/Classes/GiniConfiguration.swift
```diff
@@ -188,6 +188,15 @@ import UIKit
...
+     
+     - note: If `UIViewControllerBasedStatusBarAppearance` is set to `false` in the `Info.plist`,
+     it may not work in further versions of iOS since `UIApplication.setStatusBarStyle` method was
```
nsobadjiev: since **the** `UIApplication.setStatusBarStyle` method.

I would also consider changing "further" to "future".

GiniVision/Classes/GiniVisionUtils.swift
```diff
@@ -159,4 +159,25 @@ internal struct Colors {
...
\ No newline at end of file
+}
+
+/**
+    Set the status bar style when ViewControllerBasedStatusBarAppearance is disabled.
+ */
+
+internal func setStatusBarStyleIfNeeded(to statusBarStyle: UIStatusBarStyle,
+                                        application: UIApplication = UIApplication.shared) {
+    if !isViewControllerBasedStatusBarStyleEnabled {
```
nsobadjiev: What would happen if you don't perform this check? If it will just not apply, I might consider removing it personally.

GiniVision/Classes/GiniVisionUtils.swift
```diff
@@ -159,4 +159,25 @@ internal struct Colors {
...
     }
-}
\ No newline at end of file
+}
+
+/**
+    Set the status bar style when ViewControllerBasedStatusBarAppearance is disabled.
+ */
+
+internal func setStatusBarStyleIfNeeded(to statusBarStyle: UIStatusBarStyle,
```
nsobadjiev: I kind of hate the name `IfNeeded` in a method. Sometimes it's necessary, but I'd consider removing it here. Maybe just have the comment in a doc snippet above.

GiniVision/Classes/GiniVisionUtils.swift
kikettas: Umm, you are right. It can be confusing since you need to check the documentation to check that this method does not have effect if `View Controller based status bar` is enabled.
I just add a minor explanation above.

GiniVision/Classes/GiniVisionUtils.swift
kikettas: Given the previous comment, it does not make sense anymore to have the `IfNeeded`👍

PR #146, opened 2018-02-16
Network plugin POC

kikettas:
In order to ease the integration of the Gini Vision Library, it has been integrated all the networking within the library. 
To make the networking optional, it has been split in different `subspecs`. (`Core`, `Networking` and `Networking+Pinning`).

### How to test
Run the example app and check that is working fine without any `GiniVisionDelegate` implementation.
Now replace the `GiniVision/Networking` with `GiniVision/Networking+Pinning`on your `Podfile` and check that the pinning is possible with the new methods.

### Merging
Not merge.


Example/GiniVision/AppCoordinator.swift
```diff
@@ -46,9 +46,36 @@ final class AppCoordinator: Coordinator {
...
+        }
+        return GiniClient(clientId: "",
+                      clientSecret: "",
+                      clientEmailDomain: "")
+    }()
+    
     init(window: UIWindow) {
-        self.documentService = DocumentService()
         self.window = window
+        print("------------------------------------\n\n",
```
nsobadjiev: Why don't you use the multiline strings available in Swift 4? It would look better.

Example/GiniVision/AppCoordinator.swift
```diff
@@ -46,9 +46,36 @@ final class AppCoordinator: Coordinator {
...
+    private lazy var client: GiniClient = {[unowned self] in
+        var keys: NSDictionary?
+        if let path = self.credentialsPlistPath,
+            let keys = NSDictionary(contentsOfFile: path),
+            let client_id = keys[self.clientID] as? String,
+            let client_password = keys[self.clientPassword] as? String,
+            let client_email_domain = keys[self.clientEmailDomain] as? String,
+            !client_id.isEmpty, !client_password.isEmpty, !client_email_domain.isEmpty {
+            
+            return GiniClient(clientId: client_id,
```
nsobadjiev: I don't know about all this code - the `AppCoordinator` is already a really long file. Can't you put it in a separate file about extracting credentials?

Example/GiniVision/AppCoordinator.swift
```diff
@@ -46,9 +46,36 @@ final class AppCoordinator: Coordinator {
...
+                          clientSecret: client_password,
+                          clientEmailDomain: client_email_domain)
+        }
+        return GiniClient(clientId: "",
```
nsobadjiev: When I did the same thing for Switch, Stephan complain that returning empty strings doesn't make so much sense because it's already pretty obvious that the app wouldn't work. He suggested to throw an exception. I decided to just assert, I think. I would also consider the same thing here.

Example/GiniVision/ComponentAPICoordinator.swift
```diff
@@ -135,7 +135,7 @@ final class ComponentAPICoordinator: NSObject, Coordinator {
     }
     
     fileprivate func showResultsTableScreen(forDocument document: GINIDocument,
-                                            withResult result: GINIResult) {
+                                            withResult result: [String: GINIExtraction]) {
```
nsobadjiev: Why did you decide to remove the `GINIResult` typedef?

GiniVision/Classes/Networking/APIService.swift
```diff
@@ -0,0 +1,167 @@
...
+                         cancelationToken token: CancelationToken,
+                         completion: DocumentAnalysisCompletion?) {
+        print("🔎 Started document analysis with size \(Double(data.count) / 1024.0)")
+        
+        cancelAnalysis()
+        cancelationToken = token
+        isAnalyzing = true
+        
+        let manager = giniSDK?.documentTaskManager
+        let fileName = "fileName"
```
nsobadjiev: Why do you need a file name? And why is it always the same?

Example/GiniVision/AppCoordinator.swift
kikettas: It is still Swift 3 😅

Example/GiniVision/ComponentAPICoordinator.swift
kikettas: I consider it more confusing, since most of the times I have to check what kind of dict it is.

Example/GiniVision/AppCoordinator.swift
kikettas: It makes sense for us if we just want to test something not related to analysis results. It could lead to a sad first startup for someone that is testing the Example app, because it is quite likely that they miss adding the `plist` before. 
What do you think?

GiniVision/Classes/Networking/APIService.swift
kikettas: `createDocumentWithFilename` methods require a fileName. The `fileName` values it was just for testing purposes, It could make more sense to set it with the timestamp or uuid.

Example/GiniVision/ComponentAPICoordinator.swift
nsobadjiev: Yeah, I see your point.

One alternative is always to make a class/struct that encapsulates that information. It is probably an overkill for this case, but you might consider it if you feel confused while using these methods.

GiniVision.podspec
```diff
@@ -15,11 +15,24 @@ The Gini Vision Library for iOS provides functionality to capture documents with
   s.author           = { 'Peter Pult' => 'p.pult@gini.net' }
   s.source           = { :git => 'https://github.com/gini/gini-vision-lib-ios.git', :tag => s.version.to_s }
   s.social_media_url = 'https://twitter.com/gini'
-
   s.ios.deployment_target = '8.0'
+  s.default_subspec = 'Core'
```
nsobadjiev: I know this is what you saw in the blog post, but it's not what Cocoapods recommends. Read [this](https://guides.cocoapods.org/syntax/podspec.html#default_subspecs).

I think it's not so important and I'd say it's justified in your case, but you might want to think again. For this library, I guess you'd have to think if most people would use the core only, or networking as well.

For the pinning, I think it still shouldn't be a default.

GiniVision.podspec
```diff
@@ -15,11 +15,24 @@ The Gini Vision Library for iOS provides functionality to capture documents with
...
+
+  s.subspec 'Core' do |core|
+    core.source_files = 'GiniVision/Classes/Core/**/*'
```
nsobadjiev: I honestly prefer the way they did it in the blog post. They added the core classes in the podspec so they were automatically added to all subspecs. The only downside I see is that the "Core" subspec would be empty.

GiniVision.podspec
```diff
@@ -15,11 +15,24 @@ The Gini Vision Library for iOS provides functionality to capture documents with
...
+
+  s.subspec 'Networking' do |networking|
+    networking.source_files = 'GiniVision/Classes/Networking/*.swift', 'GiniVision/Classes/Networking/Extensions/*.swift'
+    networking.dependency "Gini-iOS-SDK", "~> 0.6.0"
+  end
 
-  s.source_files = 'GiniVision/Classes/**/*'
-  s.resources = 'GiniVision/Assets/*'
+  s.subspec 'Networking+Pinning' do |pinning|
+    pinning.source_files = 'GiniVision/Classes/Networking/Pinning/*', 'GiniVision/Classes/Networking/*.swift', 'GiniVision/Classes/Networking/Extensions/*'
```
nsobadjiev: Why are you adding the networking and networking extensions - they should be already added by the `Gini-iOS-SDK/Pinning` dependency?

GiniVision.podspec
```diff
@@ -15,11 +15,24 @@ The Gini Vision Library for iOS provides functionality to capture documents with
...
+    pinning.dependency "Gini-iOS-SDK", "~> 0.6.0"
+    pinning.dependency "Gini-iOS-SDK/Pinning", "~> 0.6.0"
```
nsobadjiev: This version (0.6.0) is repeated many times. And I assume it HAS to always be the same. Can you use a variable to make it less error prone? like `s.version` above.

nsobadjiev:
I also get this warning when I update the pods:

```
[!] Automatically assigning platform ios with version 8.0 on target GiniVision_Example because no platform was specified. Please specify a platform for this target in your Podfile. See `https://guides.cocoapods.org/syntax/podfile.html#platform`.
```

Lets put the platform in so people don't see warnings.

nsobadjiev:
I also now find the `DocumentService` a bit confusing. You have the networking part that does kind of the same thing. So you have two ways to do the same thing, which is nice, but it might be hard to understand. Lets discuss it "offline".

Example/GiniVision/ComponentAPICoordinator.swift
kikettas: For me it is more clear this way, since make it more raveled could make it more confusing for others.

GiniVision.podspec
kikettas: Exactly. I didn't make a decision yet because my first thought was what you say, having the whole package as default and then a subspec only with the UI and another with the pinning.

GiniVision.podspec
kikettas: You are right, it would much better that way.

GiniVision.podspec
kikettas: They are swift `Extensions` for the Networking, they are not part of the pinning.

GiniVision.podspec
kikettas: I cannot. At least I didn't find a way to add a custom parameter in the Podspec.

GiniVision/Classes/Networking/APIService.swift
nsobadjiev: Yeah, I see that. It might make sense to se if `createDocumentWithFilename ` really uses the file name for something meaningful and if not, to remove it. I assume that we currently don't need more than one file. But that might change with multipage.

Example/GiniVision/AppCoordinator.swift
kikettas: There will be Swift 4 before the multipage release, but I prefer to wait just before release it.

GiniVision/Classes/Networking/APIService.swift
kikettas: It is added to the query, but it will be change when we implement the multipage.

PR #154, opened 2018-02-21
Multi-page UI draft

kikettas:
This is the first draft for the Multipage UI. Although it is a draft of the multi page feature, it contains the potential UX/UI implementation. Since it is not the final version, it may contains some TODOs that will be tackled in further PRs when the multi page development has progressed in the backend.

### How to test
Run the app, take one or more pictures and try the new screen tapping on the button.

### Merging
Automatic

### Issues
#143


Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -29,11 +32,60 @@ class CameraViewControllerTests: XCTestCase {
...
+            XCTAssertTrue(self.vc.multipageReviewBackgroundView.isHidden,
+                          "multipageReviewBackgroundView should be hidden before capture the first picture")
+            self.vc.cameraDidCapture(imageData: imageData, error: nil)
+            
+            let predicate = NSPredicate(format: "isHidden == NO")
+            
+            _ = self.expectation(for: predicate,
+                                 evaluatedWith: self.vc.multipageReviewBackgroundView,
+                                 handler: .none)
+            self.waitForExpectations(timeout: 1.5, handler: { result in
```
nsobadjiev: This method has two expectations and 4 assertions. It tests a whole range of events and is therefore not a correct unit test.

Example/Tests/GINIMultipageReviewControllerTests.swift
```diff
@@ -0,0 +1,187 @@
...
+                                           cellForItemAt: IndexPath(row: 1, section: 0)) as? MultipageReviewMainCollectionCell
+        let thirdCell = vc.collectionView(vc.mainCollection,
+                                          cellForItemAt: IndexPath(row: 2, section: 0)) as? MultipageReviewMainCollectionCell
+        XCTAssertEqual(firstCell?.documentImage.image, imageDocuments[0].previewImage,
+                       "First cell image should match the one passed in the initializer")
+        XCTAssertEqual(secondCell?.documentImage.image, imageDocuments[1].previewImage,
+                       "Second cell image should match the one passed in the initializer")
+        XCTAssertEqual(thirdCell?.documentImage.image, imageDocuments[2].previewImage,
+                       "Third cell image should match the one passed in the initializer")
+        XCTAssertEqual(firstCell?.documentImage.contentMode, UIViewContentMode.scaleAspectFit,
```
nsobadjiev: Not the same test. It should be about checking if the images are correctly set, not that the image view has the correct aspect ratio. It needs to be another test.

Example/Tests/GINIMultipageReviewControllerTests.swift
```diff
@@ -0,0 +1,187 @@
...
+                       "First cell content mode should match the one passed in the initializer")
+    }
+    
+    func testMainCollectionCellSize() {
+        let vc = MultipageReviewController(imageDocuments: imageDocuments)
```
nsobadjiev: Why don't you use the view controller you already have initialized?

Example/Tests/GINIMultipageReviewControllerTests.swift
```diff
@@ -0,0 +1,187 @@
...
+                       "First cell indicator should match its position")
+        XCTAssertEqual(secondCell?.documentImage.image, imageDocuments[1].previewImage,
+                       "Second cell image should match the one passed in the initializer")
+        XCTAssertEqual(secondCell?.pageIndicator.text, "2",
+                       "Second cell indicator should match its position")
+        XCTAssertEqual(thirdCell?.documentImage.image, imageDocuments[2].previewImage,
+                       "Third cell image should match the one passed in the initializer")
+        XCTAssertEqual(thirdCell?.pageIndicator.text, "3",
+                       "Third cell indicator should match its position")
+        XCTAssertEqual(firstCell?.documentImage.contentMode, UIViewContentMode.scaleAspectFill,
```
nsobadjiev: Same thing as above - it's a different test.

Example/Tests/GINIMultipageReviewControllerTests.swift
```diff
@@ -0,0 +1,187 @@
...
+                       "First cell image should match the one passed in the initializer")
+        XCTAssertEqual(firstCell?.pageIndicator.text, "1",
+                       "First cell indicator should match its position")
+        XCTAssertEqual(secondCell?.documentImage.image, imageDocuments[1].previewImage,
+                       "Second cell image should match the one passed in the initializer")
+        XCTAssertEqual(secondCell?.pageIndicator.text, "2",
+                       "Second cell indicator should match its position")
+        XCTAssertEqual(thirdCell?.documentImage.image, imageDocuments[2].previewImage,
+                       "Third cell image should match the one passed in the initializer")
+        XCTAssertEqual(thirdCell?.pageIndicator.text, "3",
```
nsobadjiev: Here, and also above, you have the code duplicated three times. First of all, I'd only check two of them. I don't feel that the third check brings any value.

Second of all, if you're going to check all cells all the time, you should find a way to not just copy the code.

Example/Tests/GINIMultipageReviewControllerTests.swift
```diff
@@ -0,0 +1,187 @@
...
+                       "Third cell indicator should match its position")
+        XCTAssertEqual(firstCell?.documentImage.contentMode, UIViewContentMode.scaleAspectFill,
+                       "First cell content mode should match the one passed in the initializer")
+    }
+    
+    func testBottomCollectionCellSize() {
+        let vc = MultipageReviewController(imageDocuments: imageDocuments)
+        _ = vc.view
+        vc.view.setNeedsLayout()
+        vc.view.layoutIfNeeded()
```
nsobadjiev: Another duplication.

Example/Tests/GINIMultipageReviewControllerTests.swift
```diff
@@ -0,0 +1,187 @@
...
+        XCTAssertEqual(collectionInsets, vc.bottomCollectionInsets,
+                       "Main collection insets should be zero")
+    }
+    
+    func testToolBarItemsOnInitialization() {
+        XCTAssertEqual(self.vc.toolBar.items![0], self.vc.rotateButton,
+                       "First toolbar item should be the rotateButton")
+        XCTAssertEqual(self.vc.toolBar.items![2], self.vc.reorderButton,
+                       "Third toolbar item should be the rotateButton")
+        XCTAssertEqual(self.vc.toolBar.items![4], self.vc.deleteButton,
```
nsobadjiev: I think that if the item is null, an exception will be thrown and it will be a bit more difficult to follow than a plain old assertion failure (the assertion failure will print the "Fifth toolbar item should be the rotateButton" description). So, I'd personally choose to avoid risking Optional-related exceptions.

Example/Tests/GINIMultipageReviewControllerTests.swift
```diff
@@ -0,0 +1,187 @@
...
+        
+        XCTAssertEqual(collectionInsets, vc.bottomCollectionInsets,
+                       "Main collection insets should be zero")
+    }
+    
+    func testToolBarItemsOnInitialization() {
+        XCTAssertEqual(self.vc.toolBar.items![0], self.vc.rotateButton,
+                       "First toolbar item should be the rotateButton")
+        XCTAssertEqual(self.vc.toolBar.items![2], self.vc.reorderButton,
+                       "Third toolbar item should be the rotateButton")
```
nsobadjiev: You always use `rotateButton` - not reorder or delete.

Example/Tests/GINIMultipageReviewControllerTests.swift
```diff
@@ -0,0 +1,187 @@
...
+                       "toolbar and bottom collection container background colors should match")
+    }
+    
+    func testDatasourceOnDelete() {
+        let vc = MultipageReviewController(imageDocuments: imageDocuments)
+        _ = vc.view
+        vc.view.setNeedsLayout()
+        vc.view.layoutIfNeeded()
+        
+        (vc.deleteButton.customView as? UIButton)?.sendActions(for: .touchUpInside)
```
nsobadjiev: Why did you choose to send a touch event and not just invoke the Action method? Just curious...

Example/Tests/GINIMultipageReviewControllerTests.swift
```diff
@@ -0,0 +1,187 @@
...
+        
+        XCTAssertEqual(firstCell?.documentImage.image, updatedImageDocument[0].previewImage,
+                       "Second cell image should match the one passed in the initializer")
+        XCTAssertEqual(firstCell?.pageIndicator.text, "1",
+                       "First cell indicator should match its position")
+        XCTAssertEqual(secondCell?.documentImage.image, updatedImageDocument[1].previewImage,
+                       "Second cell image should match the one passed in the initializer")
+        XCTAssertEqual(secondCell?.pageIndicator.text, "2",
+                       "Second cell indicator should match its position")
+        XCTAssertEqual(thirdCell?.documentImage.image, updatedImageDocument[2].previewImage,
```
nsobadjiev: Can't you just verify the switch in one assert?

ExampleObjC/Podfile.lock
```diff
@@ -1,7 +1,19 @@
 PODS:
+<<<<<<< HEAD
```
nsobadjiev: Oooooops! Git conflict!

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -122,7 +151,8 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> Void
     fileprivate var detectedQRCodeDocument: GiniQRCodeDocument?
     fileprivate var currentQRCodePopup: QRCodeDetectedPopupView?
 
-    var didShowCamera: (() -> Void)?
+    public var didShowCamera: (() -> Void)?
+    public var didTapMultipageReviewButton: (() -> Void)?
```
nsobadjiev: I'm not sure about these closures. You have four callbacks. Maybe it would be better to have a delegate?

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -359,10 +394,7 @@ extension CameraViewController {
             do {
                 try qrDocument.validate()
                 self.showPopup(forQRDetected: qrDocument)
-            } catch let error as DocumentValidationError {
```
nsobadjiev: Why did you remove this?

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -8,6 +8,7 @@
 
 import Foundation
 
+//swiftlint:disable file_length
```
nsobadjiev: Not sure... looks like cheating.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -178,8 +182,49 @@ extension GiniScreenAPICoordinator: UINavigationControllerDelegate {
             }
         }
         
+        let isFromCameraToMultipage = (toVC == multiPageReviewController && fromVC == cameraViewController)
+        let isFromMultipageToCamera = (fromVC == multiPageReviewController && toVC == cameraViewController)
+        
+        if isFromCameraToMultipage || isFromMultipageToCamera {
```
nsobadjiev: That's ugly ;(

GiniVision/Classes/Core/GiniVisionUtils.swift
```diff
@@ -134,21 +134,29 @@ internal class Constraints {
         constraint.isActive = true
     }
     
+    class func clip(view: UIView, toSuperView superview: UIView) {
```
nsobadjiev: It's not "Clipping", is it. It's more like pinning, or filling the parent. IDK

GiniVision/Classes/Core/MultipageReviewBottomCollectionCell.swift
```diff
@@ -0,0 +1,170 @@
+//
+//  MultipageReviewBottomCollectionCell.swift
+//  GiniVision
+//
+//  Created by Enrique del Pozo Gómez on 2/1/18.
+//
+
+import Foundation
+
+final class MultipageReviewBottomCollectionCell: UICollectionViewCell {
```
nsobadjiev: I really don't understand what a bottom and a top cell is. Isn't there a better name?

GiniVision/Classes/Core/MultipageReviewBottomCollectionCell.swift
```diff
@@ -0,0 +1,170 @@
...
+    }()
+    
+    lazy var bottomContainer: UIView = {
+        let view = UIView(frame: .zero)
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.backgroundColor = .white
+        return view
+    }()
+    
+    lazy var pageIndicator: UILabel = {
```
nsobadjiev: The page indicator is a Label? I think the name is misleading - makes me think it's a page indicator. Maybe `pageIndicatorLabel`?

GiniVision/Classes/Networking/Extensions/GiniScreenAPICoordinator+Networking.swift
```diff
@@ -132,7 +132,9 @@ extension GiniScreenAPICoordinator: GiniVisionDelegate {
     
     func didCapture(document: GiniVisionDocument) {
         // Analyze document data right away with the Gini SDK for iOS to have results in as early as possible.
-        self.analyzeDocument(visionDocument: document)
+        if document.type != .image { // TODO: add analysis for images but supporting multipage.
```
nsobadjiev: What is this TODO about and why is it not resolved?

GiniVision/Classes/Core/CameraViewController.swift
kikettas: I don't like neither, actually they displease me a lot. But since we have an initializer with two closures (that I have never liked), it would suppose a breaking change for those who use the Component API.

GiniVision/Classes/Core/CameraViewController.swift
kikettas: Because it was only printing, I don't find it very useful

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
kikettas: Absolutely, but I prefer to do this instead of splitting the file or increasing the file length. If this file grows much more I will consider to split into files, since now it is readable IMO.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
kikettas: Then you had to see how it looked before 😅. This was the better I could come up with, do you know a better alternative?

GiniVision/Classes/Core/GiniVisionUtils.swift
kikettas: I doubted a lot choosing this name, and I decided to go with this because of the `clipToBounds` property. Maybe it is better the `pin` option.

GiniVision/Classes/Core/MultipageReviewBottomCollectionCell.swift
kikettas: `MultipageReviewPagesCollectionCell` sound a little bit redundant maybe, but it is more clear in case that we decide to move the collection to other place. 👍

GiniVision/Classes/Networking/Extensions/GiniScreenAPICoordinator+Networking.swift
kikettas: Basically now the analysis only works for the first page, and the backend it is not ready to test multipages.
I add the TODO so I won't forget about this.

Example/Tests/GINICameraViewControllerTests.swift
kikettas: That is what I was expecting to discuss with you. I wasn't going to use this test in the future, because for me it is more a UITest, but since the UITests aren't fixed yet I just wanted to keep it there.
But maybe I forget about this, so I will remove it.

Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -29,11 +32,60 @@ class CameraViewControllerTests: XCTestCase {
...
         
-        vc = CameraViewController(successBlock: { _ in }, failureBlock: { _ in })
         _ = vc.view
         
         XCTAssertFalse(vc.captureButton.isEnabled, "capture button should be disaled when tooltip is shown")
         
     }
+    
+    func testReviewButtonBackground() {
+        _ = vc.view
```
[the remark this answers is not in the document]
kikettas: Because there was the issue with the GiniConfiguration singleton on the tests above. I could either move it to the setUp method or move the vc initialization to the tests which need a fresh initialization.

Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -29,11 +32,60 @@ class CameraViewControllerTests: XCTestCase {
...
+        
+        let image = loadImage(withName: "invoice.jpg")
+        let imageData = UIImageJPEGRepresentation(image!, 0.9)
+        XCTAssertTrue(vc.multipageReviewBackgroundView.isHidden,
+                      "multipageReviewBackgroundView should be hidden before capture the first picture")
+        vc.cameraDidCapture(imageData: imageData, error: nil)
+        XCTAssertTrue(vc.multipageReviewBackgroundView.isHidden,
+                      "multipageReviewBackgroundView should be hidden after capture the first picture")
+        vc.cameraDidCapture(imageData: imageData, error: nil)
```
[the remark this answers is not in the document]
kikettas: Would that be worse in terms of test speed? I would have to create 3 different test.

Example/Tests/GINIMultipageReviewControllerTests.swift
kikettas: You are right!

Example/Tests/GINIMultipageReviewControllerTests.swift
kikettas: I forget to remove this when I created the separated vc. 👍

Example/Tests/GINIMultipageReviewControllerTests.swift
kikettas: I check all cells because I'm always a little bit paranoid with "if someone decides to do a check after the second cell that changes something in the next one...". What do you think?

Example/Tests/GINIMultipageReviewControllerTests.swift
kikettas: To check that the method is trigger after tapping the button. It is a double check.

Example/Tests/GINIMultipageReviewControllerTests.swift
kikettas: I want to check that the other cells also have the expected value, that is why I check every label an image.

ExampleObjC/Podfile.lock
kikettas: Forgot about the podfile in the ObjC example. I'm going to update the ObjC example app in a future PR, but thanks for pointing out! 👍

Example/Tests/GINIMultipageReviewControllerTests.swift
kikettas: Good point, I thought that it will show an optional related exceptions, instead of assert message.

Example/Tests/GINICameraViewControllerTests.swift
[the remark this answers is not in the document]
nsobadjiev: Well, it will not get better, but at least try it. I suspect that the performance impact would be negligible.

Example/Tests/GINIMultipageReviewControllerTests.swift
nsobadjiev: If that's what you prefer, I'm fine with it.

GiniVision/Classes/Core/CameraViewController.swift
nsobadjiev: Does it make sense to compile a list of such things (items that we cannot change because they would be breaking changes, but we would like to). We could actually do them when we decide to release a major version.

Alternatively, we ca start deprecating things.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
nsobadjiev: I was just complaining - transitions always look bad.

GiniVision/Classes/Networking/Extensions/GiniScreenAPICoordinator+Networking.swift
nsobadjiev: I wouldn't understand it, if I saw it in the code in 2 months. I'd write "TODO: Add multipage support".

PR #159, opened 2018-03-06
Multiple documents import

kikettas:
Multipage feature should allow importing several files from GalleryPicker/DocumentPicker. To do so, it has been implement a custom gallery to handle multiple images since the built-in one doesn't allow it.
To handle several files using _Open with/Open in_ feature, it is needed to do it outside of the library with a shared extension (an example will be provided in a different PR)
Given that CameraScreenViewController initializers should change to handle this, it has been created a delegate to do so.

Also when importing different files from the DocumentPicker, nothing happens since it has not a final decision for it.

### How to test
Run the app and pick different files from the gallery.

### Merging
Automatic

### Issues
#158


Example/Tests/AlbumsPickerViewControllerTests.swift
```diff
@@ -0,0 +1,76 @@
...
+                                               comment: "title for the albums picker view controller")
+        XCTAssertEqual(title, vc.title, "title should match the one provided in the Localizable file")
+    }
+    
+    func testNumberOfSections() {
+        XCTAssertEqual(vc.albumsTableView.numberOfSections, 1, "There should be only one section")
+    }
+    
+    func testNumberOfItems() {
+        XCTAssertEqual(vc.albumsTableView.numberOfRows(inSection: 0), 3, "There should be 4 albums")
```
nsobadjiev: 4 or 3?

Example/Tests/AlbumsPickerViewControllerTests.swift
```diff
@@ -0,0 +1,76 @@
...
+                       "table view cell heght should match AlbumsPickerTableViewCell height")
+    }
+    
+    func testTableCellSelection() {
+        let delegate = AlbumsPickerViewControllerDelegateMock()
+        vc.delegate = delegate
+        
+        let selectedIndex = IndexPath(row: 0, section: 0)
+        let selectedAlbum = galleryManager.albums[selectedIndex.row]
+        vc.tableView(vc.albumsTableView, didSelectRowAt: selectedIndex)
```
nsobadjiev: Wouldn't it be better if you just selected the cell.
I guess you don't have to test that the `didSelectRowAt ` is called, but it's possible that the table view has selection disabled or something. It might be a better way to test it.

Example/Tests/AlbumsPickerViewControllerTests.swift
```diff
@@ -0,0 +1,76 @@
...
+        XCTAssertEqual(selectedAlbum, delegate.selectedAlbum,
+                       "selected album should match the one delivered to the delegate")
+    }
+    
+    func testTableCellContent() {
+        let firstIndex = IndexPath(row: 0, section: 0)
+        let secondIndex = IndexPath(row: 1, section: 0)
```
nsobadjiev: You are testing two things in a single test again

Example/Tests/GalleryCoordinatorTests.swift
```diff
@@ -0,0 +1,132 @@
...
+        coordinator = GalleryCoordinator(galleryManager: self.galleryManager,
+                                         giniConfiguration: GiniConfiguration.sharedConfiguration)
+    }
+    
+    func testGalleryCoordinatorStart() {
+        coordinator.start()
+        
+        XCTAssertTrue(galleryManager.isCaching, "gallery manager should have started caching")
+        XCTAssertEqual(coordinator.galleryNavigator.viewControllers.count, 2,
+                       "there should be 2 view controller in the navigator when the gallery coordinator starts")
```
nsobadjiev: view controller**s** (plural)

Example/Tests/GalleryCoordinatorTests.swift
```diff
@@ -0,0 +1,132 @@
...
+            XCTAssertTrue(self.coordinator.selectedImageDocuments.isEmpty,
+                          "selected image documents collection should be cleared after cancelling")
+        }
+    }
+    
+    func testOpenImages() {
+        let delegate = GalleryCoordinatorDelegateMock()
+        coordinator.delegate = delegate
+        
+        selectImage(at: 0, in: galleryManager.albums[2]) { _ in
```
nsobadjiev: Isn't there a way to have the select image method not work async? I would try to avoid that when possible

Example/Tests/ImagePickerViewControllerTests.swift
```diff
@@ -0,0 +1,72 @@
...
+        _ = vc.view
+    }
+    
+    func testViewControllerTitle() {
+        let title = currentAlbum.title
+        XCTAssertEqual(title, vc.title, "view controller title should match the album title")
+    }
+    
+    func testNumberOfSections() {
+        XCTAssertEqual(vc.collectionView.numberOfSections, 1, "There should be only one section")
```
nsobadjiev: Either have the assertion descriptions start with a capital letter or a lowercase one. But be consistent.

Example/Tests/ImagePickerViewControllerTests.swift
```diff
@@ -0,0 +1,72 @@
...
+    }
+    
+    func testNumberOfItems() {
+        XCTAssertEqual(vc.collectionView.numberOfItems(inSection: 0), 2,
+                       "There should be only 1 image in the first album")
```
nsobadjiev: 1 or 2?

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -177,23 +184,36 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> Void
     
     /**
      Designated initializer for the `CameraViewController` which allows
+     to set the `GiniConfiguration for the camera screen`.
+     All the interactions with this screen are handle by `CameraViewControllerDelegate`.
```
nsobadjiev: handle**d** by

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -228,6 +248,16 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> Void
         NotificationCenter.default.removeObserver(self)
     }
     
+    fileprivate func didPick(validatedDocuments documents: [GiniVisionDocument]) {
+        if let delegate = delegate {
+            delegate.camera(self, didCaptureDocuments: documents)
+        } else if let firstDocument = documents.first {
+            successBlock?(firstDocument)
+        } else {
+            print("It has not been specified a CameraViewControllerDelegate")
```
nsobadjiev: "A CameraViewControllerDelegate has not been specified".

Also, isn't it time to have a better logging solution? It's a library after all.

GiniVision/Classes/Core/ContainerNavigationController.swift
```diff
@@ -8,10 +8,15 @@
 
 import UIKit
 
+/**
+ Container that wraps a UINavigationController in order to handle rotation.
+ Parent coordinator should be always `nil` excepts when there is no possibility
```
nsobadjiev: **The** parent coordinator should be always `nil` except<strike>s</strike> when there is no possibility

GiniVision/Classes/Core/Extensions/Collection.swift
```diff
@@ -8,10 +8,34 @@
 
 import Foundation
 
+extension Collection {
+    var isNotEmpty: Bool {
```
nsobadjiev: What's the point? You don't want to have double negatives?

GiniVision/Classes/Core/Extensions/Collection.swift
```diff
@@ -8,10 +8,34 @@
...
 internal extension Collection where Iterator.Element == CFString {
     
     var strings: [ String ] {
         return self.map { $0 as String }
     }
     
 }
+
+extension Collection where Iterator.Element == GiniVisionDocument {
+    var isAssorted: Bool {
```
nsobadjiev: So Assorted means that all elements have the same document type?
How about calling it a "homogenous document type"?
Also, can't you make it simpler? Why not get the type of the first one and then filter/reduce to the ones who don't have the same type. If the array is not empty, the collection is "assorted".

GiniVision/Classes/Core/FilePickerManager.swift
```diff
@@ -24,18 +26,21 @@ internal final class FilePickerManager: NSObject {
         }
     }
     
+    override init() {
+        super.init()
+        DispatchQueue.global().async {
+            self.galleryCoordinator.start()
```
nsobadjiev: I don't like having an async operation in a constructor. A constructor would not have side effects.
And I fear, that's not what people would expect from a class.

GiniVision/Classes/Core/FilePickerManager.swift
```diff
@@ -57,70 +66,51 @@ internal final class FilePickerManager: NSObject {
...
-        if let pickedImage = info[UIImagePickerControllerOriginalImage] as? UIImage,
-            let imageData = UIImageJPEGRepresentation(pickedImage, 1.0) {
-            filePicked(withData: imageData)
-        }
+extension FilePickerManager: GalleryCoordinatorDelegate {
+    func gallery(_ coordinator: GalleryCoordinator, didSelectImageDocuments imageDocuments: [GiniImageDocument]) {
+        didPickDocuments(imageDocuments)
         
-        picker.dismiss(animated: true, completion: nil)
+        coordinator.rootViewController.dismiss(animated: true, completion: nil)
```
nsobadjiev: I'm not sure about this. You expose some coordinator details. That it has a root controller and that it presented something. I'd rather have a method that tells the coordinator that the picker is no longer needed.

GiniVision/Classes/Core/FilePickerManager.swift
```diff
@@ -57,70 +66,51 @@ internal final class FilePickerManager: NSObject {
...
     }
     
-    func imagePickerControllerDidCancel(_ picker: UIImagePickerController) {
-        picker.dismiss(animated: true, completion: nil)
+    func gallery(_ coordinator: GalleryCoordinator, didCancel: Void) {
+        coordinator.rootViewController.dismiss(animated: true, completion: nil)
```
nsobadjiev: same here.

GiniVision/Classes/Core/Gallery/AlbumsPickerTableViewCell.swift
```diff
@@ -0,0 +1,116 @@
...
+final class AlbumsPickerTableViewCell: UITableViewCell {
+    
+    static let identifier = "AlbumsPickerTableViewCellIdentifier"
+    static let height: CGFloat = 90.0
+    let padding = UIEdgeInsets(top: 10, left: 10, bottom: 10, right: 10)
+    
+    lazy var albumAccesoryView: UIButton = {
+        let button = UIButton(frame: .zero)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.setImage(UIImageNamedPreferred(named: "disclosureIndicator"), for: .normal)
```
nsobadjiev: Does it make sense to replace these strings with an enum (not in this PR).

GiniVision/Classes/Core/Gallery/ImagePickerViewController.swift
```diff
@@ -0,0 +1,137 @@
...
+    }
+    
+    // MARK: - Others
+ 
+    fileprivate func scrollToBottomOnStartup() {
+        // This tweak is needed to fix an issue with the UICollectionView. UICollectionView doesn't
+        // scroll to the bottom on `viewWillAppear`, which is right after `viewDidLayoutSubviews`.
+        // Since this method can be called several times during the lifecycle, there should be
+        // a one-time scrolling before the view appears for the first time.
+        if !isInitialized {
```
nsobadjiev: what is you used `Dispatch.once`or whatever the name was?

GiniVision/Classes/Core/GiniVision.swift
```diff
@@ -116,16 +116,40 @@ import UIKit
      - note: Screen API only.
      
      - parameter delegate: An instance conforming to the `GiniVisionDelegate` protocol.
-     - parameter importedDocument:  A document which comes from a source different than CameraViewController.
+     - parameter importedDocuments: Documents which come from a source different than `CameraViewController`.
```
nsobadjiev: Documents **that** sounds better

Example/Tests/AlbumsPickerViewControllerTests.swift
kikettas: `didSelectRowAt` method is not triggered when you select a cell programmatically unfortunately. UIKit rules.

Example/Tests/GalleryCoordinatorTests.swift
kikettas: Unfortunately the image is loaded asynchronously and I have to wait for it. Although I don't like async operations in unit tests, I think this way is at least clear.

Example/Tests/ImagePickerViewControllerTests.swift
kikettas: You are right 👍

GiniVision/Classes/Core/CameraViewController.swift
kikettas: I thought about it when adding TrustKit to the other SDK but I forgot to add it to Github. Thanks for reminding me about it!

GiniVision/Classes/Core/Extensions/Collection.swift
kikettas: To make it more clear in the code, instead of having the likely missable `!` before of `isEmpty`

GiniVision/Classes/Core/Extensions/Collection.swift
kikettas: I didn't like neither _Mixed_ nor _Diversed_, so I decided to go with _Assorted_. But how would you call that long `homogenous document type` variable? That is what I wanted to express but IMO is clear with one word. What do you think?

kikettas: Like that way of getting the bool btw

GiniVision/Classes/Core/Gallery/AlbumsPickerTableViewCell.swift
kikettas: Really good idea, added to the Github issues.

GiniVision/Classes/Core/Gallery/ImagePickerViewController.swift
kikettas: `Dispatch
The free function dispatch_once is no longer available in Swift. In Swift, you can use lazily initialized globals or static properties and get the same thread-safety and called-once guarantees as dispatch_once provided.`

GiniVision/Classes/Core/FilePickerManager.swift
kikettas: The `rootViewController` is always exposed since is part of the `Coordinator` protocol. So it cannot be hidden.
Do you have a better idea?

GiniVision/Classes/Core/FilePickerManager.swift
kikettas: Answered above.

Example/Tests/AlbumsPickerViewControllerTests.swift
nsobadjiev: Oh... yeah, now I remember. I've had the same problem. It's intended behaviour.
It actually makes sense to a degree. For most use cases (that are not testing), if you select a cell in code, you wouldn't want to trigger the action for it.

Example/Tests/GalleryCoordinatorTests.swift
nsobadjiev: Ideally, you'd want to avoid triggering such "extra features" in a unit test. You don't need the image to load so the unit test shouldn't trigger it. But that's in an ideal world. If it's a lot of effort to not load the images, keep it as is.

GiniVision/Classes/Core/FilePickerManager.swift
nsobadjiev: That's not the point. It's fine to have the root controller if it support another use case.
The problem I have is that you also expose (and "hardcode") that the coordinator presents view controllers (as opposed to pushing them or embedding them). You also mess with the class internals. It's the age-old discussion between asking a class and telling a class what to do.

So, I'd prefer to have "Tell the coordinator to remove the view controller", rather than "Ask the coordinator for its root controller and tell that root controller to dismiss its presented controller".

Example/Tests/GalleryCoordinatorTests.swift
kikettas: The only way is to set an empty value in the dictionary and assign it later once the image has been fetched. IMO this is quite raveled and could lead to big issues. In this case I prefer to keep it as it is.

GiniVision/Classes/Core/FilePickerManager.swift
kikettas: Got it now. Yes, it makes sense what you say.

GiniVision/Classes/Core/Extensions/Collection.swift
nsobadjiev: I personally don't see a big benefit, but if you like it, ok.

GiniVision/Classes/Core/FilePickerManager.swift
nsobadjiev: What about this one?

GiniVision/Classes/Core/Extensions/Collection.swift
```diff
@@ -8,10 +8,27 @@
...
```
nsobadjiev: I just feel that Assorted doesn't translate well into English. At least I don't associate it with a mixture of different typed things.

I'd use something simple like:
* `containsDifferentTypes`
* `mixedDocumentTypes`

Idk, I also cannot find a well-suited name

kikettas: Yeah, maybe make sense to be more explicit in this case since this extension might be public in a future PR. I will go with the `containsDifferentTypes `

PR #173, opened 2018-03-09
Handle multiple type collections

kikettas:
When importing documents from either external apps or the Document picker, there could be the case that both PDFs and images are imported at the same time. Since multipage doesn't support this behavior yet, the library won't handle it and it will show a message when this happens.

Although the Gini Vision Library support multiple types imported from other apps, it doesn't mean that iOS is able to forward several files. As described [here](https://stackoverflow.com/questions/47203396/import-multiple-files-in-app-via-ios11-files-app/47415226#47415226), before iOS there was an option to read the Inbox folder and check for several files, even if the AppDelegate only forward one `URL`. 
Nevertheless, this feature can be used from other views outside of the library but inside of the current application.

### How to test
Run the app and try to import several files from the DocumentPicker.

### Merging
Automatic

Issue #169

### DIFF
--- a/Example/Tests/GiniScreenAPICoordinatorTests.swift
+++ b/Example/Tests/GiniScreenAPICoordinatorTests.swift
@@ -1,0 +1,130 @@
+//
+//  GiniScreenAPICoordinatorTests.swift
+//  GiniVision_Tests
+//
+//  Created by Enrique del Pozo Gómez on 3/8/18.
+//  Copyright © 2018 Gini GmbH. All rights reserved.
+//
+
+import XCTest
+@testable import GiniVision
+
+final class GiniScreenAPICoordinatorTests: XCTestCase {
+    
+    var coordinator: GiniScreenAPICoordinator!
+    let giniConfiguration = GiniConfiguration()
+    let delegate = GiniVisionDelegateMock()
+    
+    override func setUp() {
+        super.setUp()
+        giniConfiguration.openWithEnabled = true
+        coordinator = GiniScreenAPICoordinator(withDelegate: delegate, giniConfiguration: giniConfiguration)
+    }
+    
+    func testNavControllerCountAfterStartWithoutDocuments() {
+        let rootViewController = coordinator.start(withDocuments: nil)
+        _ = rootViewController.view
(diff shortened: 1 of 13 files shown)


Example/Tests/GiniScreenAPICoordinatorTests.swift
```diff
@@ -0,0 +1,119 @@
...
+        _ = rootViewController.view
+        let screenNavigator = rootViewController.childViewControllers.first as? UINavigationController
+        
+        XCTAssertNotNil(screenNavigator?.viewControllers.first as? CameraViewController,
+                        "first view controller is not a CameraViewController")
+        XCTAssertNotNil(screenNavigator?.viewControllers.last as? MultipageReviewController,
+                        "last view controller is not a MultipageReviewController")
+    }
+    
+    func testNavControllerCountAfterStartWithAPDF() {
```
nsobadjiev: You don't really need the "A" in "a PDF".

Example/Tests/GiniScreenAPICoordinatorTests.swift
```diff
@@ -0,0 +1,119 @@
...
+    }
+    
+    func testNavControllerTypesAfterStartWithImages() {
+        let capturedImages = [loadImageDocument(withName: "invoice"), loadImageDocument(withName: "invoice2")]
+
+        let rootViewController = coordinator.start(withDocuments: capturedImages)
+        _ = rootViewController.view
+        let screenNavigator = rootViewController.childViewControllers.first as? UINavigationController
+        
+        XCTAssertNotNil(screenNavigator?.viewControllers.first as? CameraViewController,
```
nsobadjiev: I'm not sure about checking a view controller that's not even visible in screen. It's a conceptual thing, but usually it's better to test only the desired result of what you're doing. Not some side effects. In other words, you should tests your classes kind of as black boxes. Test only the end effect, not the way it is achieved.

We have a lot of work to do to reach that level, but I just wanted to make a note of it here. It seemed to me that some of these tests are too detailed.

kikettas: But then in this specific case, how do you test that both screens are in the navigation stack?

PR #176, opened 2018-03-12
Multipage feature toggle

kikettas:
Since the multipage feature will be optional, it has been added a toggle to enable/disable it along with different check in all multipage related screens.

### How to test
Run the app and try to use the multipage feature. Go to Setting from the API selection screen and disable the multipage feature. After that try to use the multipage and see that it is not possible.

### Merging
Automatic

Issue #172

### DIFF
--- a/Example/GiniVision/SettingsViewController.swift
+++ b/Example/GiniVision/SettingsViewController.swift
@@ -22,6 +22,7 @@ final class SettingsViewController: UIViewController {
     @IBOutlet weak var fileImportControl: UISegmentedControl!
     @IBOutlet weak var openWithSwitch: UISwitch!
     @IBOutlet weak var qrCodeScanningSwitch: UISwitch!
+    @IBOutlet weak var multipageSwitch: UISwitch!
     @IBAction func fileImportOptions(_ sender: UISegmentedControl) {
         switch sender.selectedSegmentIndex {
         case 0:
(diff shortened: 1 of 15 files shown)


Example/GiniVision/SettingsViewController.swift
```diff
@@ -49,11 +50,19 @@ final class SettingsViewController: UIViewController {
...
+        delegate?.settings(settingViewController: self, didChangeConfiguration: giniConfiguration)
+
+    }
+    
     override func viewDidLoad() {
         super.viewDidLoad()
         openWithSwitch.setOn(giniConfiguration.openWithEnabled, animated: true)
         qrCodeScanningSwitch.setOn(giniConfiguration.qrCodeScanningEnabled, animated: true)
-
+        multipageSwitch.setOn(giniConfiguration.multipageEnabled, animated: true)
```
nsobadjiev: It makes no sense to have it animated in the `viewDidLoad` method.

Example/Tests/GalleryCoordinatorTests.swift
```diff
@@ -12,6 +12,7 @@ import XCTest
 final class GalleryCoordinatorTests: XCTestCase {
     
     let galleryManager = GalleryManagerMock()
+    let giniConfiguration: GiniConfiguration = GiniConfiguration.sharedConfiguration
```
nsobadjiev: You keep using the shared configuration in tests, but you leave it in undefined state after the test is run. It can lead to subtle bug, especially since the order of tests is not guaranteed.

I'd suggest not using the shared configuration. Or at the very least, reset the values in the `tearDown`.

Example/GiniVision/SettingsViewController.swift
kikettas: True 👍

PR #180, opened 2018-03-20
Adapt API SDK changes

kikettas:
The API SDK was updated with several changes (Bolts and Cancellation token) that has to be reflected in the GVL.

### How to test
Run the app and try to analyze a document.

### Merging
Automatic

Issue #178

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -1,17 +1,17 @@
 PODS:
-  - Bolts (1.1.5):
-    - Bolts/AppLinks (= 1.1.5)
-    - Bolts/Tasks (= 1.1.5)
-  - Bolts/AppLinks (1.1.5):
+  - Bolts (1.9.0):
+    - Bolts/AppLinks (= 1.9.0)
+    - Bolts/Tasks (= 1.9.0)
+  - Bolts/AppLinks (1.9.0):
     - Bolts/Tasks
-  - Bolts/Tasks (1.1.5)
+  - Bolts/Tasks (1.9.0)
   - Gini-iOS-SDK (0.6.0):
     - Gini-iOS-SDK/Core (= 0.6.0)
   - Gini-iOS-SDK/Core (0.6.0):
-    - Bolts (~> 1.1.5)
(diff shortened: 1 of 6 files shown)


Example/Podfile.lock
```diff
@@ -1,27 +1,27 @@
...
-  - Bolts/Tasks (1.1.5)
-  - Gini-iOS-SDK (0.6.0):
-    - Gini-iOS-SDK/Core (= 0.6.0)
-  - Gini-iOS-SDK/Core (0.6.0):
-    - Bolts (~> 1.1.5)
-  - Gini-iOS-SDK/Pinning (0.6.0):
-    - Bolts (~> 1.1.5)
-    - TrustKit (~> 1.5.2)
+  - Bolts/Tasks (1.9.0)
+  - Gini-iOS-SDK (1.0.0-alpha.1):
```
nsobadjiev: Merging this into `multipage` is fine, but just remember to **not** have an alpha version when you go to `develop`.

GiniVision/Classes/Networking/APIService.swift
```diff
@@ -170,58 +171,57 @@ final class APIService: APIServiceProtocol {
...
-        return { [weak self] (task: BFTask?) -> Any! in
-            guard let `self` = self else { return nil }
-            
-            if let token = token, token.cancelled {
-                return BFTask.cancelled()
-            }
-            if task?.error != nil {
-                return self.giniSDK?.sessionManager.logIn()
+    fileprivate func getSessionBlock(cancelationToken token: BFCancellationToken? = nil)
+        -> ((BFTask<AnyObject>) -> Any?) {
```
nsobadjiev: This return value scares me. Why didn't you continue using `BFContinuationBlock` and since it's about logging in, can't you replace the `Any` and `AnyObject` with something more specific?

kikettas: Because it breaks the compiler. For some reason it cannot infer the `ResultType`, even if you specified it.

Issue #164, opened 2018-03-06
Reestructure project


kikettas:
Since there are more and more classes, it is needed to put them into folders.

PR #181, opened 2018-03-22
Document picker error handling

kikettas:
When importing pictures from the pickers (gallery, explorer or dragndrop), there could be errors that are thrown outside of the `CameraViewController` or `DocumentPickerCoordinator`, for example the documents limit exceeded error.

### How to test
Run the app and try to import more than 10 images, both from gallery or files explorer.

### Merging
Automatic

### DIFF
--- a/Example/Tests/GINICameraViewControllerTests.swift
+++ b/Example/Tests/GINICameraViewControllerTests.swift
@@ -10,10 +10,15 @@ final class CameraViewControllerTests: XCTestCase {
         let imageData = UIImageJPEGRepresentation(image!, 0.9)!
         return imageData
     }()
+    var giniConfiguration: GiniConfiguration = GiniConfiguration.sharedConfiguration
+    lazy var screenAPICoordinator = GiniScreenAPICoordinator(withDelegate: nil,
+                                                             giniConfiguration: self.giniConfiguration)
     
     override func setUp() {
         super.setUp()
-        vc = CameraViewController(successBlock: { _ in }, failureBlock: { _ in })
+        giniConfiguration.multipageEnabled = true
+        vc = CameraViewController(giniConfiguration: giniConfiguration)
+        vc.delegate = screenAPICoordinator
     }
     
     func testInitialization() {
@@ -22,7 +27,8 @@ final class CameraViewControllerTests: XCTestCase {
     
     func testTooltipWhenFileImportDisabled() {
         ToolTipView.shouldShowFileImportToolTip = true
-        GiniConfiguration.sharedConfiguration.fileImportSupportedTypes = .none
+        giniConfiguration.fileImportSupportedTypes = .none
+        vc = CameraViewController(giniConfiguration: giniConfiguration)
         _ = vc.view
         
         XCTAssertNil(vc.toolTipView, "ToolTipView should not be created when file import is disabled.")
@@ -36,6 +42,7 @@ final class CameraViewControllerTests: XCTestCase {
         // Disable onboarding on launch
         GiniConfiguration.sharedConfiguration.onboardingShowAtLaunch = false
         GiniConfiguration.sharedConfiguration.onboardingShowAtFirstLaunch = false
+        vc = CameraViewController(giniConfiguration: giniConfiguration)
         
         _ = vc.view
         
@@ -69,5 +76,81 @@ final class CameraViewControllerTests: XCTestCase {
                       "multipageReviewBackgroundView should not be hidden after capture the second picture")
         
     }
+    
+    func testPickerCompletionBlockWhenNoErrors() {
+        let documents = [GiniImageDocument(data: imageData, imageSource: .external)]
+        let expect = expectation(description: "Document validation finishes")
+
+        vc.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, didDismiss in
+            expect.fulfill()
+            XCTAssertNil(error, "Completion block should not return an error when pciking one image")
+        }
+        
+        waitForExpectations(timeout: 2, handler: nil)
+    }
+    
+    func testPickerCompletionBlockWhenNoErrorsWithDeprecatedInit() {
+        let documents = [loadImageDocument(withName: "invoice")]
+        let expect = expectation(description: "Document validation finishes")
+        vc = CameraViewController(success: {_ in}, failure: {_ in})
+        vc.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, didDismiss in
+            expect.fulfill()
+            XCTAssertNil(error, "Completion block should not return an error when using the deprecated initializer")
+        }
+        
+        waitForExpectations(timeout: 2, handler: nil)
+    }
+    
+    func testPickerCompletionBlockWhenFailedPDF() {
+        let failedPDF = loadPDFDocument(withName: "testPDF")
+        failedPDF.error = DocumentValidationError.pdfPageLengthExceeded
+        let documents = [failedPDF]
+        let expect = expectation(description: "Document validation finishes")
+        
+        vc.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, didDismiss in
+            expect.fulfill()
+            XCTAssertNil(error, "Completion block should not return an error form outside when it is a PDF")
+        }
+        
+        waitForExpectations(timeout: 2, handler: nil)
+    }
+    
+    func testPickerCompletionBlockWhenFailedImageAndMultipageDisabled() {
+        let failedImage = loadImageDocument(withName: "invoice")
+        failedImage.error = DocumentValidationError.imageFormatNotValid
+        let documents = [failedImage]
+        
+        giniConfiguration.multipageEnabled = false
+        vc = CameraViewController(giniConfiguration: giniConfiguration)
+        vc.delegate = screenAPICoordinator
+        
+        let expect = expectation(description: "Document validation finishes")
+
+        vc.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, didDismiss in
+            expect.fulfill()
+            XCTAssertNil(error, "When multipage is disabled there should not be an error coming from outside of the camera screen")
+        }
+        
+        waitForExpectations(timeout: 2, handler: nil)
+    }
+    
+    func testPickerCompletionBlockWhenTooManyPages() {
+        var documents: [GiniVisionDocument] = []
+        for _ in 0...GiniPDFDocument.maxPagesCount {
+            documents.append(loadImageDocument(withName: "invoice"))
+        }
+
+        let expect = expectation(description: "Document validation finishes")
+        vc.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, didDismiss in
+            expect.fulfill()
+            let error = error as? DocumentValidationError
+            XCTAssertTrue(error == DocumentValidationError.filesPickedCountExceeded,
+                          "Completion block should return the filesPickedCounteExceedede error from outside of the caera screen")
+        }
+        
+        waitForExpectations(timeout: 2, handler: nil)
+    }
+    
+    
 }
 
--- a/Example/Tests/GINICameraViewControllerTests.swift
+++ b/Example/Tests/GINICameraViewControllerTests.swift
@@ -37,11 +37,11 @@ final class CameraViewControllerTests: XCTestCase {
     
     func testCaptureButtonDisabledWhenToolTipIsShown() {
         ToolTipView.shouldShowFileImportToolTip = true
-        GiniConfiguration.sharedConfiguration.fileImportSupportedTypes = .pdf_and_images
+        giniConfiguration.fileImportSupportedTypes = .pdf_and_images
         
         // Disable onboarding on launch
-        GiniConfiguration.sharedConfiguration.onboardingShowAtLaunch = false
-        GiniConfiguration.sharedConfiguration.onboardingShowAtFirstLaunch = false
+        giniConfiguration.onboardingShowAtLaunch = false
+        giniConfiguration.onboardingShowAtFirstLaunch = false
         vc = CameraViewController(giniConfiguration: giniConfiguration)
         
         _ = vc.view
@@ -143,8 +143,8 @@ final class CameraViewControllerTests: XCTestCase {
         let expect = expectation(description: "Document validation finishes")
         vc.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, didDismiss in
             expect.fulfill()
-            let error = error as? DocumentValidationError
-            XCTAssertTrue(error == DocumentValidationError.filesPickedCountExceeded,
+            let error = error as? FilePickerError
+            XCTAssertTrue(error == FilePickerError.filesPickedCountExceeded,
                           "Completion block should return the filesPickedCounteExceedede error from outside of the caera screen")
         }
         
--- a/Example/Tests/GINICameraViewControllerTests.swift
+++ b/Example/Tests/GINICameraViewControllerTests.swift
@@ -104,39 +104,6 @@ final class CameraViewControllerTests: XCTestCase {
         waitForExpectations(timeout: 2, handler: nil)
     }
     
-    func testPickerCompletionBlockWhenFailedPDF() {
-        let failedPDF = loadPDFDocument(withName: "testPDF")
-        failedPDF.error = DocumentValidationError.pdfPageLengthExceeded
-        let documents = [failedPDF]
-        let expect = expectation(description: "Document validation finishes")
-        
-        cameraViewController.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, _ in
-            expect.fulfill()
-            XCTAssertNil(error, "Completion block should not return an error from outside when it is a PDF")
-        }
-        
-        waitForExpectations(timeout: 2, handler: nil)
-    }
-    
-    func testPickerCompletionBlockWhenFailedImageAndMultipageDisabled() {
-        let failedImage = loadImageDocument(withName: "invoice")
-        failedImage.error = DocumentValidationError.imageFormatNotValid
-        let documents = [failedImage]
-        
-        giniConfiguration.multipageEnabled = false
-        cameraViewController = CameraViewController(giniConfiguration: giniConfiguration)
-        cameraViewController.delegate = screenAPICoordinator
-        
-        let expect = expectation(description: "Document validation finishes")
-
-        cameraViewController.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, _ in
-            expect.fulfill()
-            XCTAssertNil(error, "When multipage is disabled there should not be an error coming from outside of the camera screen")
-        }
-        
-        waitForExpectations(timeout: 2, handler: nil)
-    }
-    
     func testPickerCompletionBlockWhenTooManyPages() {
         var documents: [GiniVisionDocument] = []
         for _ in 0...GiniPDFDocument.maxPagesCount {
--- a/Example/Tests/GINICameraViewControllerTests.swift
+++ b/Example/Tests/GINICameraViewControllerTests.swift
@@ -4,34 +4,37 @@ import AVFoundation
 
 final class CameraViewControllerTests: XCTestCase {
     
-    var vc: CameraViewController!
+    var cameraViewController: CameraViewController!
+    var giniConfiguration: GiniConfiguration!
+    var screenAPICoordinator: GiniScreenAPICoordinator!
     lazy var imageData: Data = {
         let image = self.loadImage(withName: "invoice.jpg")
         let imageData = UIImageJPEGRepresentation(image!, 0.9)!
         return imageData
     }()
-    var giniConfiguration: GiniConfiguration = GiniConfiguration.sharedConfiguration
-    lazy var screenAPICoordinator = GiniScreenAPICoordinator(withDelegate: nil,
-                                                             giniConfiguration: self.giniConfiguration)
+
     
     override func setUp() {
         super.setUp()
+        giniConfiguration = GiniConfiguration.sharedConfiguration
         giniConfiguration.multipageEnabled = true
-        vc = CameraViewController(giniConfiguration: giniConfiguration)
-        vc.delegate = screenAPICoordinator
+        cameraViewController = CameraViewController(giniConfiguration: giniConfiguration)
+        screenAPICoordinator = GiniScreenAPICoordinator(withDelegate: nil,
+                                                        giniConfiguration: self.giniConfiguration)
+        cameraViewController.delegate = screenAPICoordinator
     }
     
     func testInitialization() {
-        XCTAssertNotNil(vc, "view controller should not be nil")
+        XCTAssertNotNil(cameraViewController, "view controller should not be nil")
     }
     
     func testTooltipWhenFileImportDisabled() {
         ToolTipView.shouldShowFileImportToolTip = true
         giniConfiguration.fileImportSupportedTypes = .none
-        vc = CameraViewController(giniConfiguration: giniConfiguration)
-        _ = vc.view
+        cameraViewController = CameraViewController(giniConfiguration: giniConfiguration)
+        _ = cameraViewController.view
         
-        XCTAssertNil(vc.toolTipView, "ToolTipView should not be created when file import is disabled.")
+        XCTAssertNil(cameraViewController.toolTipView, "ToolTipView should not be created when file import is disabled.")
         
     }
     
@@ -42,46 +45,46 @@ final class CameraViewControllerTests: XCTestCase {
         // Disable onboarding on launch
         giniConfiguration.onboardingShowAtLaunch = false
         giniConfiguration.onboardingShowAtFirstLaunch = false
-        vc = CameraViewController(giniConfiguration: giniConfiguration)
+        cameraViewController = CameraViewController(giniConfiguration: giniConfiguration)
         
-        _ = vc.view
+        _ = cameraViewController.view
         
-        XCTAssertFalse(vc.captureButton.isEnabled, "capture button should be disaled when tooltip is shown")
+        XCTAssertFalse(cameraViewController.captureButton.isEnabled, "capture button should be disaled when tooltip is shown")
         
     }
     
     func testReviewButtonBackgroundBeforeCapturing() {
-        _ = vc.view
+        _ = cameraViewController.view
 
-        XCTAssertTrue(vc.multipageReviewBackgroundView.isHidden,
+        XCTAssertTrue(cameraViewController.multipageReviewBackgroundView.isHidden,
                       "multipageReviewBackgroundView should be hidden before capture the first picture")
         
     }
     
     func testReviewButtonBackgroundAfter1ImageWasCaptured() {
-        _ = vc.view
+        _ = cameraViewController.view
 
-        vc.cameraDidCapture(imageData: imageData, error: nil)
-        XCTAssertTrue(vc.multipageReviewBackgroundView.isHidden,
+        cameraViewController.cameraDidCapture(imageData: imageData, error: nil)
+        XCTAssertTrue(cameraViewController.multipageReviewBackgroundView.isHidden,
                       "multipageReviewBackgroundView should be hidden after capture the first picture")
         
     }
     
     func testReviewButtonBackgroundAfter2ImagesWereCaptured() {
-        _ = vc.view
+        _ = cameraViewController.view
 
-        vc.cameraDidCapture(imageData: imageData, error: nil)
-        vc.cameraDidCapture(imageData: imageData, error: nil)
-        XCTAssertTrue(vc.multipageReviewBackgroundView.isHidden,
+        cameraViewController.cameraDidCapture(imageData: imageData, error: nil)
+        cameraViewController.cameraDidCapture(imageData: imageData, error: nil)
+        XCTAssertTrue(cameraViewController.multipageReviewBackgroundView.isHidden,
                       "multipageReviewBackgroundView should not be hidden after capture the second picture")
         
     }
     
-    func testPickerCompletionBlockWhenNoErrors() {
+    func testPickerCompletionBlockWhenNoErrorsOccurred() {
         let documents = [GiniImageDocument(data: imageData, imageSource: .external)]
         let expect = expectation(description: "Document validation finishes")
 
-        vc.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, _ in
+        cameraViewController.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, _ in
             expect.fulfill()
             XCTAssertNil(error, "Completion block should not return an error when pciking one image")
         }
@@ -89,11 +92,11 @@ final class CameraViewControllerTests: XCTestCase {
         waitForExpectations(timeout: 2, handler: nil)
     }
     
-    func testPickerCompletionBlockWhenNoErrorsWithDeprecatedInit() {
+    func testPickerCompletionBlockWhenNoErrorsOccurredWithDeprecatedInit() {
         let documents = [loadImageDocument(withName: "invoice")]
         let expect = expectation(description: "Document validation finishes")
-        vc = CameraViewController(success: {_ in}, failure: {_ in})
-        vc.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, _ in
+        cameraViewController = CameraViewController(success: {_ in}, failure: {_ in})
+        cameraViewController.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, _ in
             expect.fulfill()
             XCTAssertNil(error, "Completion block should not return an error when using the deprecated initializer")
         }
@@ -107,9 +110,9 @@ final class CameraViewControllerTests: XCTestCase {
         let documents = [failedPDF]
         let expect = expectation(description: "Document validation finishes")
         
-        vc.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, _ in
+        cameraViewController.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, _ in
             expect.fulfill()
-            XCTAssertNil(error, "Completion block should not return an error form outside when it is a PDF")
+            XCTAssertNil(error, "Completion block should not return an error from outside when it is a PDF")
         }
         
         waitForExpectations(timeout: 2, handler: nil)
@@ -121,12 +124,12 @@ final class CameraViewControllerTests: XCTestCase {
         let documents = [failedImage]
         
         giniConfiguration.multipageEnabled = false
-        vc = CameraViewController(giniConfiguration: giniConfiguration)
-        vc.delegate = screenAPICoordinator
+        cameraViewController = CameraViewController(giniConfiguration: giniConfiguration)
+        cameraViewController.delegate = screenAPICoordinator
         
         let expect = expectation(description: "Document validation finishes")
 
-        vc.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, _ in
+        cameraViewController.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, _ in
             expect.fulfill()
             XCTAssertNil(error, "When multipage is disabled there should not be an error coming from outside of the camera screen")
         }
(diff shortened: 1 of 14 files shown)


Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -10,10 +10,15 @@ final class CameraViewControllerTests: XCTestCase {
...
     }()
+    var giniConfiguration: GiniConfiguration = GiniConfiguration.sharedConfiguration
+    lazy var screenAPICoordinator = GiniScreenAPICoordinator(withDelegate: nil,
+                                                             giniConfiguration: self.giniConfiguration)
     
     override func setUp() {
         super.setUp()
-        vc = CameraViewController(successBlock: { _ in }, failureBlock: { _ in })
+        giniConfiguration.multipageEnabled = true
+        vc = CameraViewController(giniConfiguration: giniConfiguration)
```
nsobadjiev: Is there a reason why the `GiniConfiguration` and `screenAPICoordinator` are initialized when they are declared, but the `CameraViewController` is created in he setup method?

Just a question of consistency.

Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -22,7 +27,8 @@ final class CameraViewControllerTests: XCTestCase {
     
     func testTooltipWhenFileImportDisabled() {
         ToolTipView.shouldShowFileImportToolTip = true
-        GiniConfiguration.sharedConfiguration.fileImportSupportedTypes = .none
+        giniConfiguration.fileImportSupportedTypes = .none
+        vc = CameraViewController(giniConfiguration: giniConfiguration)
```
nsobadjiev: It's a bit confusing that you have a property called `vc` that's already initialized with a view controller and you are reassigning it with a new one.

Secondly, `vc` is not a good name ;)

Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -31,11 +37,12 @@ final class CameraViewControllerTests: XCTestCase {
...
         ToolTipView.shouldShowFileImportToolTip = true
-        GiniConfiguration.sharedConfiguration.fileImportSupportedTypes = .pdf_and_images
+        giniConfiguration.fileImportSupportedTypes = .pdf_and_images
         
         // Disable onboarding on launch
-        GiniConfiguration.sharedConfiguration.onboardingShowAtLaunch = false
-        GiniConfiguration.sharedConfiguration.onboardingShowAtFirstLaunch = false
+        giniConfiguration.onboardingShowAtLaunch = false
+        giniConfiguration.onboardingShowAtFirstLaunch = false
+        vc = CameraViewController(giniConfiguration: giniConfiguration)
```
nsobadjiev: Same here.

Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -69,5 +76,81 @@ final class CameraViewControllerTests: XCTestCase {
                       "multipageReviewBackgroundView should not be hidden after capture the second picture")
         
     }
+    
+    func testPickerCompletionBlockWhenNoErrors() {
```
nsobadjiev: Either `WithNoErrors` or `WhenNoErrorsOccurred`. Otherwise it's not a correct sentence. Also applies for other tests in this class.

Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -69,5 +76,81 @@ final class CameraViewControllerTests: XCTestCase {
...
+    
+    func testPickerCompletionBlockWhenFailedPDF() {
+        let failedPDF = loadPDFDocument(withName: "testPDF")
+        failedPDF.error = DocumentValidationError.pdfPageLengthExceeded
+        let documents = [failedPDF]
+        let expect = expectation(description: "Document validation finishes")
+        
+        vc.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, _ in
+            expect.fulfill()
+            XCTAssertNil(error, "Completion block should not return an error form outside when it is a PDF")
```
nsobadjiev: _from*_

Also, I don't understand - why not return an error?

Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -69,5 +76,81 @@ final class CameraViewControllerTests: XCTestCase {
...
+        for _ in 0...GiniPDFDocument.maxPagesCount {
+            documents.append(loadImageDocument(withName: "invoice"))
+        }
+
+        let expect = expectation(description: "Document validation finishes")
+        vc.documentPicker(DocumentPickerCoordinator(), didPick: documents, from: .gallery) { error, _ in
+            expect.fulfill()
+            let error = error as? FilePickerError
+            XCTAssertTrue(error == FilePickerError.filesPickedCountExceeded,
+                          "Completion block should return the filesPickedCounteExceedede error from outside of the caera screen")
```
nsobadjiev: _ FilePickerError.filesPickedCountExceeded*_

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -11,9 +10,38 @@ import UIKit
 import AVFoundation
 
 @objc public protocol CameraViewControllerDelegate: class {
-    @objc func camera(_ viewController: CameraViewController, didCaptureDocuments documents: [GiniVisionDocument])
-    @objc func camera(_ viewController: CameraViewController, didFailCaptureWithError error: CameraError)
+    /**
+     Called when a user take a picture, import a PDF/QRCode or import one or several images.
```
nsobadjiev: take**s**, import**s**, import**s**

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -11,9 +10,38 @@ import UIKit
...
+     Once that the method has been implemented, it is necessary to check if the number of
```
nsobadjiev: Once <strike>that</strike> the method is implemented

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -11,9 +10,38 @@ import UIKit
...
+     documents accumulated doesn't exceed the minimun (`GiniImageDocument.maxPagesCount`).
+     Afterwards, it is mandatory to call the `DocumentPickerCompletion` block passing the error
+     `CameraError.filesPickedCountExceeded` (if the page count limit was exceeded) and
+     the inner completion block that will be executed once the gallery dismissal animation finishes.
```
nsobadjiev: "Completes" is a good synonym for "finishes".

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -11,9 +10,38 @@ import UIKit
 import AVFoundation
 
 @objc public protocol CameraViewControllerDelegate: class {
-    @objc func camera(_ viewController: CameraViewController, didCaptureDocuments documents: [GiniVisionDocument])
-    @objc func camera(_ viewController: CameraViewController, didFailCaptureWithError error: CameraError)
+    /**
```
nsobadjiev: Why didn't you format the comment like a javadoc or something?

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -11,9 +10,38 @@ import UIKit
...
+    
+    /**
+     Called when the `CameraViewController` appears.
+     
+     - parameter viewController: Camera view controller that appears.
+     */
     @objc func cameraDidAppear(_ viewController: CameraViewController)
+    
+    /**
+     Called when a user tap the `MultipageReviewButton` (the one with the thumbnail of the images/s taken).
```
nsobadjiev: tap**s**
"image/images", or better yet "image(s)".

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -251,11 +279,16 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> Void
...
+    fileprivate func didPick(validatedDocuments documents: [GiniVisionDocument],
+                             completion: DocumentPickerCompletion?) {
         if let delegate = delegate {
-            delegate.camera(self, didCaptureDocuments: documents)
+            delegate.camera(self, didCaptureDocuments: documents, completion: completion)
         } else if let firstDocument = documents.first {
+            completion?(nil, nil)
             successBlock?(firstDocument)
+            if documents.count > 1 {
+                // TODO: Show a warning in the logger
```
nsobadjiev: We don't like TODOs in a finished PR ;)

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -251,11 +279,16 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> Void
...
+            }
         } else {
             print("It has not been specified a CameraViewControllerDelegate")
```
nsobadjiev: a `CameraViewControllerDelegate ` has not been specified.
This sounds waaay better. In fact, yours is probably not even correct English.

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -463,13 +496,17 @@ extension CameraViewController {
...
         
         if giniConfiguration.multipageEnabled {
-            self.animateToControlsView(imageDocument: imageDocument) {
-                self.didPick(validatedDocuments: [imageDocument])
+            self.didPick(validatedDocuments: [imageDocument]) { error, _ in
+                guard let error = error else {
+                    self.animateToControlsView(imageDocument: imageDocument)
+                    return
+                }
+                self.showErrorDialog(for: error)
```
nsobadjiev: Maybe this is just me, but I find this `guard` confusing. I'm used to seeing the edge cases within the guard and the successful case after it. Here, it's reversed.

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -614,15 +651,53 @@ extension CameraViewController {
...
+// MARK: - DocumentPickerCoordinatorDelegate
+
+extension CameraViewController: DocumentPickerCoordinatorDelegate {
+
+    func documentPicker(_ coordinator: DocumentPickerCoordinator,
+                        didPick documents: [GiniVisionDocument],
+                        from: DocumentPickerType,
+                        completion: DocumentPickerCompletion?) {
+        self.validate(importedDocuments: documents) { validatedDocuments in
+            if let firstElement = validatedDocuments.first,
```
nsobadjiev: Why only check the first one?

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -614,15 +651,53 @@ extension CameraViewController {
...
 
+// MARK: - DocumentPickerCoordinatorDelegate
+
+extension CameraViewController: DocumentPickerCoordinatorDelegate {
+
+    func documentPicker(_ coordinator: DocumentPickerCoordinator,
+                        didPick documents: [GiniVisionDocument],
+                        from: DocumentPickerType,
+                        completion: DocumentPickerCompletion?) {
+        self.validate(importedDocuments: documents) { validatedDocuments in
```
nsobadjiev: This closure is huge!! I don't even want to try understanding it.

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -633,81 +708,52 @@ extension CameraViewController {
         }
     }
     
-    fileprivate func validate(importedDocuments documents: [GiniVisionDocument]) {
+    fileprivate func validate(importedDocuments documents: [GiniVisionDocument],
+                              completion: @escaping ([GiniVisionDocument]) -> Void) {
         let loadingView = addValidationLoadingView()
```
nsobadjiev: You are working with views in a `validate` method??? Not good!

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -633,81 +708,52 @@ extension CameraViewController {
...
+                completion(validatedDocuments)
             }
+            
         }
     }
     
-    fileprivate func process(validatedImportedDocuments documents: [GiniVisionDocument]) {
+    fileprivate func process(validatedImportedDocuments documents: [GiniVisionDocument],
+                             completion: DocumentPickerCompletion?) {
+        let didValidated: DocumentPickerCompletion
```
nsobadjiev: _didValidate*_

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -779,34 +821,44 @@ extension CameraViewController {
...
+
+        let dialog = errorDialog(withMessage: message,
+                                 cancelActionTitle: cancelActionTitle,
+                                 confirmActionTitle: confirmActionTitle,
+                                 confirmAction: possitiveAction ?? confirmAction)
         
-        present(alertViewController, animated: true, completion: nil)
+        present(dialog, animated: true, completion: nil)
     }
 }
```
nsobadjiev: Line 864... Please split it or something.

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -779,34 +821,44 @@ extension CameraViewController {
...
-        let alertViewController = UIAlertController(title: nil, message: message, preferredStyle: .alert)
-        alertViewController.addAction(UIAlertAction(title: "Abbrechen", style: .cancel, handler: { _ in
-            alertViewController.dismiss(animated: true, completion: nil)
-        }))
-        alertViewController.addAction(UIAlertAction(title: "Andere Datei wählen", style: .default, handler: { _ in
-            self.showImportFileSheet()
-        }))
+    func showErrorDialog(for error: Error, possitiveAction: (() -> Void)? = nil) {
+        let message: String
+        var cancelActionTitle: String = "Abbrechen"
```
nsobadjiev: Not localized

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -779,34 +821,44 @@ extension CameraViewController {
...
+        var confirmActionTitle: String? = "Andere Datei wählen"
```
nsobadjiev: Not localized.

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
```diff
@@ -10,11 +10,39 @@ import Foundation
 import MobileCoreServices
 import Photos
 
-internal final class FilePickerManager: NSObject {
+protocol DocumentPickerCoordinatorDelegate: class {
+    /**
+     Called when a user pick one or several files from both the gallery or files explorer.
```
nsobadjiev: pick**s**
either - both implies that they used both of them, not one of them.

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
```diff
@@ -10,11 +10,39 @@ import Foundation
...
+     The completion might provide with errors than must be handled here before dismissing the
```
nsobadjiev: The completion might provide <strike>with</strike> errors tha**t** must be handled here before dismissing the

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
```diff
@@ -10,11 +10,39 @@ import Foundation
...
+     pickers. It only applies for the `GalleryCoordinator` since on one side it is not possible
```
nsobadjiev: If you have "on one side", it's nice to also have "on the other side"

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
```diff
@@ -67,8 +98,29 @@ internal final class FilePickerManager: NSObject {
...
+    }
+    
+    func showErrorDialog(for error: Error, from viewController: UIViewController) {
+        let dialog: UIAlertController
+        
+        switch error {
+        case let error as FilePickerError where error == .photoLibraryAccessDenied:
+            dialog = errorDialog(withMessage: error.message,
+                                 cancelActionTitle: "Abbrechen",
+                                 confirmActionTitle: "Zugriff erteilen",
```
nsobadjiev: Both strings are not localized.

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
```diff
@@ -67,8 +98,29 @@ internal final class FilePickerManager: NSObject {
...
+                                 confirmAction: UIApplication.shared.openAppSettings)
+        case let error as FilePickerError where error == .filesPickedCountExceeded:
 
-        from.present(documentPicker, animated: true, completion: nil)
+            dialog = errorDialog(withMessage: error.message,
+                                 cancelActionTitle: "OK")
```
nsobadjiev: Not localized.

GiniVision/Classes/Core/GiniError.swift
```diff
@@ -26,6 +28,10 @@ public protocol GiniVisionError: Error {}
     /// Capturing could not be completed.
     case captureFailed
     
+    public var message: String {
+        // TODO: Add localized string for each case
```
nsobadjiev: Then do it.

GiniVision/Classes/Core/GiniError.swift
```diff
@@ -36,6 +42,10 @@ public protocol GiniVisionError: Error {}
     /// Unknown error during review.
     case unknown
     
+    public var message: String {
```
nsobadjiev: The same property is present in the Camera error. Can you combine them? It's quite common that an error has a message/description.

GiniVision/Classes/Core/GiniError.swift
```diff
@@ -36,6 +42,10 @@ public protocol GiniVisionError: Error {}
...
+        // TODO: Add localized string for each case
```
nsobadjiev: Then do it.

GiniVision/Classes/Core/GiniError.swift
```diff
@@ -46,7 +56,30 @@ public protocol GiniVisionError: Error {}
     
     /// Camera roll can not be loaded because the user has denied authorization in the past.
     case photoLibraryAccessDenied
+    
+    /// Number of files picked exceeded
```
nsobadjiev: Exceeded what?
"/// Max Number of files picked exceeded"

That also applies for the variable name. It could be more descriptive. For instance, `maxFilesCountExceeded `

GiniVision/Classes/Core/GiniError.swift
```diff
@@ -46,7 +56,30 @@ public protocol GiniVisionError: Error {}
...
+    case filesPickedCountExceeded
+    
+    /// Mixed documents unsupported
+    case mixedDocumentsUnsupported
 
+    public var message: String {
```
nsobadjiev: Again, a `message` property.

GiniVision/Classes/Core/GiniImageDocument.swift
```diff
@@ -21,6 +21,7 @@ final public class GiniImageDocument: NSObject, GiniVisionDocument {
     public var previewImage: UIImage?
     public var isReviewable: Bool
     public var isImported: Bool
+    public var error: Error?
```
nsobadjiev: I cannot totally agree with the semantics here. It's convenient to have the error with the document. But semantically, a document is just a document. If it has an error, it cannot be constructed.

I would **seriously** consider moving it somewhere else.

GiniVision/Classes/Core/GiniPDFDocument.swift
```diff
@@ -18,8 +18,8 @@ final public class GiniPDFDocument: NSObject, GiniVisionDocument {
     public var previewImage: UIImage?
     public var isReviewable: Bool
     public var isImported: Bool
+    public var error: Error?
```
nsobadjiev: Same here.

GiniVision/Classes/Core/GiniPDFDocument.swift
```diff
@@ -64,7 +64,7 @@ final public class GiniPDFDocument: NSObject, GiniVisionDocument {
     
     public func checkType() throws {
         if self.data.isPDF {
-            if case 1...MAX_PDF_PAGES_COUNT = self.numberPages {
+            if case 1...GiniPDFDocument.maxPagesCount = self.numberPages {
```
nsobadjiev: Same thing. If Why would a Document know about the 10 page constraint. a 1000 page PDF is still a valid PDF document. This constraint comes later, in another business logic layer.

GiniVision/Classes/Core/GiniQRCodeDocument.swift
```diff
@@ -17,6 +17,7 @@ import Foundation
     }()
     public var isReviewable: Bool = false
     public var isImported: Bool = false
+    public var error: Error?
```
nsobadjiev: Again, this error... It's also duplicated now quite a few times. I know the error is just a protocol and this declaration has to be repeated here, but... it's by not 5 times.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -235,45 +235,53 @@ extension GiniScreenAPICoordinator: UINavigationControllerDelegate {
 // MARK: - Camera Screen
 
 extension GiniScreenAPICoordinator: CameraViewControllerDelegate {
-    func camera(_ viewController: CameraViewController, didCaptureDocuments documents: [GiniVisionDocument]) {
+    func camera(_ viewController: CameraViewController,
```
nsobadjiev: This method is so long and covers so many cases that it gives me a headache.

Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -10,10 +10,15 @@ final class CameraViewControllerTests: XCTestCase {
         let imageData = UIImageJPEGRepresentation(image!, 0.9)!
         return imageData
     }()
+    var giniConfiguration: GiniConfiguration = GiniConfiguration.sharedConfiguration
```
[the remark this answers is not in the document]
kikettas: The problem is always the same. Although now I pass the `GiniConfiguration`, the singleton is spread all over the code and I can't control some of its usages.
The idea is to remove the singleton in a future PR and _then_, create a new instance in every test.

Example/Tests/GINICameraViewControllerTests.swift
kikettas: The `screenAPICoordinator` is initialized when it was declared because I don't need another instance of it, making the tests slightly faster, although I guess that It is no saving a lot of time. I can move it to the initialization method.

Example/Tests/GINICameraViewControllerTests.swift
kikettas: I do this because the `giniConfiguration` can be used in the initialization phase. 
Yes, it can have a better name.

Example/Tests/GINICameraViewControllerTests.swift
kikettas: Answered above.

Example/Tests/GINICameraViewControllerTests.swift
kikettas: Because when there is a PDF, there are no validation from outside (since it can only be one document not several documents.)

GiniVision/Classes/Core/CameraViewController.swift
kikettas: But I prefer to have this before releasing the new version and implemented it when I create the Logger. Otherwise I will forget about this line.

GiniVision/Classes/Core/CameraViewController.swift
kikettas: Because the 'positive' case here is to have an error, since the completion block is focused in the error.

GiniVision/Classes/Core/CameraViewController.swift
kikettas: Nice catch, this is a bug. It should be first `firstElementWithError`

GiniVision/Classes/Core/CameraViewController.swift
kikettas: 🤦‍♂️

GiniVision/Classes/Core/CameraViewController.swift
kikettas: It was there from previous implementation where it made sense, but now as you say, it is not good! 👍

GiniVision/Classes/Core/CameraViewController.swift
kikettas: #164

GiniVision/Classes/Core/CameraViewController.swift
kikettas: I didn't want to do this in this PR, but if it ok for you 🤷‍♂️

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
kikettas: For some reason I forgot to finish the explanation 🤦‍♂️

GiniVision/Classes/Core/GiniError.swift
kikettas: As I said before I didn't want to make a lot of changes in this PR, but if it is ok for you then I will do it.

GiniVision/Classes/Core/GiniImageDocument.swift
kikettas: This was the last option for me. I *really* don't like having and error inside of a document, since as you well say, it should not be part of a document because a document is a document.
But the problem was that in the `CameraViewControllerDelegate` I have to send the documents back along with theirs errors (If they had).
Firstly I tried with a `[GiniVisionDocument: GiniError]` dictionary but `GiniVisionDocument` does not and can not implement `Hashable`. Secondly I tried the easy and ugly approach, using a tuple but it is not supported by ObjC.
The other option that I ruled out was to have a Wrapper or something similar, but in my opinion it can be more confusing for those who use the Component API.

GiniVision/Classes/Core/GiniPDFDocument.swift
kikettas: Answered above.

GiniVision/Classes/Core/GiniQRCodeDocument.swift
kikettas: But I don't get this, it is part of a protocol as it is `type` or `data`.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
kikettas: Yes it is. I thought about splitting it but it can be more confusing for those who implement the Component API.

GiniVision/Classes/Core/GiniPDFDocument.swift
kikettas: It is part of their own validation check.

GiniVision/Classes/Core/GiniError.swift
kikettas: Answered above.

GiniVision/Classes/Core/GiniImageDocument.swift
nsobadjiev: Still not a good reason. This thing will come back to haunt you. It's an architectural error that you will probably never fix. Even when you decide to roll out a new major version, this thing will be so deeply integrated that you will not dare to touch it.

GiniVision/Classes/Core/GiniPDFDocument.swift
nsobadjiev: What does that mean? Whose's validation check?

GiniVision/Classes/Core/GiniQRCodeDocument.swift
nsobadjiev: It's not a big deal. I just don't like properties in a protocol because you just have to repeat the same thing in all classes implementing it.

GiniVision/Classes/Core/GiniImageDocument.swift
kikettas: Do you have a better idea then? Because I know that is an architectural error but I don't see a better approach.

GiniVision/Classes/Core/GiniPDFDocument.swift
kikettas: Each document has different kinds of validation.

GiniVision/Classes/Core/GiniError.swift
nsobadjiev: What's up with this?

GiniVision/Classes/Core/GiniQRCodeDocument.swift
kikettas: I don't see it as an issue, since you conform to a common interface/protocol. The idea would be to add it only in the `GiniVisionDocument`, but it doesn't allow stored properties (there is a workaround to add stored properties to a protocol, but I would rather not to use it).

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
kikettas: They would have to implement several methods that would lead to the same target, process a document.

Or you mean just our implementation? Then it would be possible to split into different methods.

GiniVision/Classes/Core/GiniError.swift
kikettas: It is answered above.

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -11,9 +10,38 @@ import UIKit
...
+     Once the method has been implemented, it is necessary to check if the number of
+     documents accumulated doesn't exceed the minimun (`GiniImageDocument.maxPagesCount`).
+     Afterwards, it is mandatory to call the `DocumentPickerCompletion` block passing the error
+     `CameraError.maxFilesPickedCountExceeded` (if the page count limit was exceeded) and
+     the inner completion block that will be executed once the gallery dismissal animation completes.
+     
+     - parameter viewController: `CameraViewController` where the documents were taken.
+     - parameter documents: One or several documents either captured or imported in
+     the `CameraViewController`. They can contain an error produced in the validation process.
+     - parameter completion: `DocumentPickerCompletion` block used to check if there is an issue with
```
nsobadjiev: But then it is not exactly a completion handler, is it? The whole delegate method is about completion, but the closure is about validation. That is if I understand correctly.

GiniVision/Classes/Core/CameraViewController.swift
nsobadjiev: Ok, that makes sense. But why is the completion handler focused on the error. It also handles the successful case, doesn't it? That means that if you change the naming, it becomes a standard completion closure. 

It's not important, but I feel like it's similar to a `isNotEmpty` method, which requires a bit more mental power to understand than `isEmpty`.

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -614,15 +651,56 @@ extension CameraViewController {
...
+extension CameraViewController: DocumentPickerCoordinatorDelegate {
+
+    func documentPicker(_ coordinator: DocumentPickerCoordinator,
+                        didPick documents: [GiniVisionDocument],
+                        from: DocumentPickerType,
+                        completion: DocumentPickerCompletion?) {
+        let loadingView = addValidationLoadingView()
+        self.validate(importedDocuments: documents) { validatedDocuments in
+            loadingView.removeFromSuperview()
+            let elementsWithError = validatedDocuments.filter { $0.1 != nil}
```
nsobadjiev: Can you use names, rather than 0 and 1? Without some domain knowledge it's difficult to understand what that is.

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -614,15 +651,56 @@ extension CameraViewController {
...
+            if let firstElement = elementsWithError.first,
```
nsobadjiev: Why only the first error?

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -614,15 +651,56 @@ extension CameraViewController {
...
+                let error = firstElement.1,
+                (!self.giniConfiguration.multipageEnabled || firstElement.0.type != .image) {
```
nsobadjiev: Why is that? I really don't understand? Why only if the type is not an image and there's no multipage?

Also, it's so many conditions that my head hurts.

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
```diff
@@ -10,11 +10,39 @@ import Foundation
...
+     picker has been dismissed when there are no errors.
+     */
+    func documentPicker(_ coordinator: DocumentPickerCoordinator,
+                        didPick documents: [GiniVisionDocument],
+                        from picker: DocumentPickerType,
+                        completion: DocumentPickerCompletion?)
+}
+
+public typealias DidDismissPickerCompletion = () -> Void
+public typealias DocumentPickerCompletion = (Error?, DidDismissPickerCompletion?) -> Void
```
nsobadjiev: Completion within completion? It's getting out of hand!

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -614,15 +651,56 @@ extension CameraViewController {
...
+                if let completion = completion {
+                    completion(nil) {
+                        self.showErrorDialog(for: error)
+                    }
+                } else {
+                    self.showErrorDialog(for: error)
+                }
+            } else {
```
nsobadjiev: This is a good place to **not** have this huge closure, but separate it logically. Here, you handle both the error and processing case. Just have two methods.
If there was an error -> `processError`
Else -> `processDocuments`

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -614,15 +651,56 @@ extension CameraViewController {
...
+                self.process(validatedImportedDocuments: validatedDocuments.map {$0.0 }) { [weak self] error, didDismiss in
+                    guard let `self` = self else { return }
+                    // This is needed since the `UIDocumentPickerViewController` is automatically
+                    // dismissed and the drag&drop is done in this view controller.
```
nsobadjiev: This comment confuses me even more.

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -614,15 +651,56 @@ extension CameraViewController {
...
+                    if from != .gallery {
+                        guard let error = error else {
+                            didDismiss?()
+                            return
+                        }
+                        
+                        self.showErrorDialog(for: error)
+                    } else {
+                        completion?(error, didDismiss)
```
nsobadjiev: Look at this (the else statement) and the if statement proceeding it. There are so many cases. That's why I complained that you have a completion handler within a completion handler. It's a rabbit hole. And it's endless.

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -779,34 +822,47 @@ extension CameraViewController {
...
-    fileprivate func showNotValidDocumentError(message: String) {
-        
-        let alertViewController = UIAlertController(title: nil, message: message, preferredStyle: .alert)
-        alertViewController.addAction(UIAlertAction(title: "Abbrechen", style: .cancel, handler: { _ in
-            alertViewController.dismiss(animated: true, completion: nil)
-        }))
-        alertViewController.addAction(UIAlertAction(title: "Andere Datei wählen", style: .default, handler: { _ in
-            self.showImportFileSheet()
-        }))
+    func showErrorDialog(for error: Error, possitiveAction: (() -> Void)? = nil) {
```
nsobadjiev: Why is this still in the `CameraViewController`? The file is so long. The last thing it needs is 60 lines of error messages.

nsobadjiev: But ok, there's a ticket for it.

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
```diff
@@ -10,11 +10,39 @@ import Foundation
 import MobileCoreServices
 import Photos
 
-internal final class FilePickerManager: NSObject {
+protocol DocumentPickerCoordinatorDelegate: class {
+    /**
+     Called when a user picks one or several files from either the gallery or files explorer.
```
nsobadjiev: or **the** files explorer

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
```diff
@@ -10,11 +10,39 @@ import Foundation
...
+     The completion might provide errors that must be handled here before dismissing the
+     pickers. It only applies for the `GalleryCoordinator` since on one side it is not possible
```
nsobadjiev: applies to

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
```diff
@@ -67,8 +98,31 @@ internal final class FilePickerManager: NSObject {
         if !device.isIpad {
             setStatusBarStyle(to: .default)
         }
-
-        from.present(documentPicker, animated: true, completion: nil)
+        
+        viewController.present(documentPicker, animated: true, completion: nil)
+    }
+    
+    func showErrorDialog(for error: Error, from viewController: UIViewController) {
```
nsobadjiev: You really need a new class/component that creates these error dialogs. Almost the same code is spread all over the code.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
nsobadjiev: yeah, just the implementation.

GiniVision/Classes/Core/CameraViewController.swift
kikettas: Because this only applies for PDFs and QR codes, or when the multipage is not enabled. (The error is shown right away)

GiniVision/Classes/Core/CameraViewController.swift
kikettas: I explained it above. But yes, too many conditions. I will try to simplify it.

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
kikettas: The naming can be improved.

GiniVision/Classes/Core/CameraViewController.swift
kikettas: I simplified it a little bit and also renamed the closure to make them more clear.

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
kikettas: I already extracted the logic of creating an alertviewcontroller to another class (`errorDialog` method), so know I only need to specify what is the message and the actions.

PR #189, opened 2018-03-29
Restrict pdfs when image(s) have been already selected

kikettas:
PDF file import should be disabled if there is already an image imported/taken

### How to test
Run the app, take a picture and try to import a PDF from the Files explorer.

### Merging
Automatic

### DIFF
--- a/GiniVision/Classes/Core/CameraViewController.swift
+++ b/GiniVision/Classes/Core/CameraViewController.swift
@@ -545,6 +545,7 @@ extension CameraViewController {
         multipageReviewBackgroundView.isHidden = !showingStack
         multipageReviewButton.setImage(image, for: .normal)
         multipageReviewButton.isUserInteractionEnabled = image != nil
+        filePickerManager.isPDFSelectionAllowed = image == nil
     }
     
     func updatePreviewViewOrientation() {
--- a/GiniVision/Classes/Core/CameraViewController.swift
+++ b/GiniVision/Classes/Core/CameraViewController.swift
@@ -784,7 +784,7 @@ extension CameraViewController {
         }
         
         alertViewController.addAction(UIAlertAction(title: "Dokumente", style: .default) { [unowned self] _ in
-            self.documentPickerCoordinator.isPDFSelectionAllowed = self.multipageReviewButton.image(for: .normal) == nil
+            self.documentPickerCoordinator.isPDFSelectionAllowed = !self.imagesAlreadyPicked()
             self.documentPickerCoordinator.showDocumentPicker(from: self)
         })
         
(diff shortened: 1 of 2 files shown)


GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -784,6 +784,7 @@ extension CameraViewController {
         }
         
         alertViewController.addAction(UIAlertAction(title: "Dokumente", style: .default) { [unowned self] _ in
+            self.documentPickerCoordinator.isPDFSelectionAllowed = self.multipageReviewButton.image(for: .normal) == nil
```
a-szotyori: I would move `self.multipageReviewButton.image(for: .normal) == nil` into a dedicated method for clarity.

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
```diff
@@ -227,10 +234,10 @@ extension DocumentPickerCoordinator: UIDropInteractionDelegate {
         }
     }
     
-    fileprivate func loadDocuments<T: NSItemProviderReading>(ofClass classs: T.Type,
-                                                             from session: UIDropSession,
-                                                             in group: DispatchGroup,
-                                                             completion: @escaping (([T]?) -> Void)) {
+    private func loadDocuments<T: NSItemProviderReading>(ofClass classs: T.Type,
```
a-szotyori: Why not `fileprivate` anymore?

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
```diff
@@ -190,13 +193,17 @@ extension DocumentPickerCoordinator: UIDocumentPickerDelegate {
 @available(iOS 11.0, *)
 extension DocumentPickerCoordinator: UIDropInteractionDelegate {
     func dropInteraction(_ interaction: UIDropInteraction, canHandle session: UIDropSession) -> Bool {
-        let isItemsSelectionAllowed = session.items.count > 1 ? giniConfiguration.multipageEnabled : true
+        if !isPDFDropSelectionAllowed(forSession: session) {
```
a-szotyori: Groot?
```
guard isPDFDropSelectionAllowed(forSession: session) else {
    return false;
}
```

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
kikettas: Because I want to make it visible only for the extension (Drag and drop related stuff).

GiniVision/Classes/Core/DocumentPickerCoordinator.swift
kikettas: Nice catch!

PR #190, opened 2018-04-04
Fix gallery issues

kikettas:
There was a couple of issues related to gallery:

- Opening the gallery before giving access to it resulted in a blank collection since images weren't cached after that.
- When fetching assets from the OS, if the asset was in iCloud it returned `nil`.

### How to test
Make a clean install of the app and try to open the gallery after granting permissions to it. Also select an asset from iCloud.

### Merging
Automatic

### DIFF
--- a/GiniVision/Classes/Core/Gallery/GalleryCoordinator.swift
+++ b/GiniVision/Classes/Core/Gallery/GalleryCoordinator.swift
@@ -204,8 +204,8 @@ extension GalleryCoordinator: ImagePickerViewControllerDelegate {
                         if let data = data {
                             DispatchQueue.main.async {
                                 viewController.removeFromDownloadingItems(index: index)
+                                viewController.selectCell(at: index)
                             }
-                            viewController.selectCell(at: index)
                             self.addSelected(asset, withData: data)
                         }
                     }
@@ -216,9 +216,11 @@ extension GalleryCoordinator: ImagePickerViewControllerDelegate {
     }
     
     func imagePicker(_ viewController: ImagePickerViewController,
-                     didDeselectAsset asset: Asset) {
-        if let index = selectedImageDocuments.index(forKey: asset.identifier) {
-            selectedImageDocuments.remove(at: index)
+                     didDeselectAsset asset: Asset,
+                     at index: IndexPath) {
+        if let documentIndex = selectedImageDocuments.index(forKey: asset.identifier) {
+            viewController.deselectCell(at: index)
+            selectedImageDocuments.remove(at: documentIndex)
         }
         
         if let selectedItems = viewController.collectionView.indexPathsForSelectedItems, selectedItems.isEmpty {
--- a/GiniVision/Classes/Core/Gallery/GalleryCoordinator.swift
+++ b/GiniVision/Classes/Core/Gallery/GalleryCoordinator.swift
@@ -10,8 +10,7 @@ import Photos
 
 protocol GalleryCoordinatorDelegate: class {
     func gallery(_ coordinator: GalleryCoordinator,
-                 didSelectImageDocuments imageDocuments: [GiniImageDocument],
-                 completion: @escaping () -> Void)
+                 didSelectImageDocuments imageDocuments: [GiniImageDocument])
     func gallery(_ coordinator: GalleryCoordinator, didCancel: Void)
 }
 
@@ -103,6 +102,13 @@ final class GalleryCoordinator: NSObject, Coordinator {
     
     func dismissGallery(completion: (() -> Void)? = nil) {
         rootViewController.dismiss(animated: true, completion: completion)
+        resetToInitialState()
+    }
+    
+    private func resetToInitialState() {
+        self.selectedImageDocuments.removeAll()
+        self.currentImagePickerViewController?.deselectAllCells()
+        self.currentImagePickerViewController?.navigationItem.setRightBarButton(cancelButton, animated: true)
     }
     
     // MARK: - Bar button actions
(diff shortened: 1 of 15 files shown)


GiniVision/Classes/Core/Gallery/GalleryManager.swift
```diff
@@ -45,12 +50,19 @@ final class GalleryManager: GalleryManagerProtocol {
         }
     }
     
-    func fetchImageData(from asset: Asset, completion: @escaping ((Data) -> Void)) {
-        cachingImageManager.requestImageData(for: asset.value, options: nil) { data, _, _, _ in
-            if let data = data {
-                completion(data)
-            }
+    func fetchImageData(from asset: Asset, isRemote: Bool, completion: @escaping ((Data?) -> Void)) {
```
a-szotyori: throw new FlagArgumentDetectedException("The flag argument isRemote signals that this method does more than one thing. Consider moving the two execution branches into separate methods.")

I think fetchRemoteImageData() is a suitable name for the isRemote: true case.

GiniVision/Classes/Core/Gallery/GalleryManager.swift
```diff
@@ -90,7 +102,7 @@ extension GalleryManager {
         return assets
     }
     
-    fileprivate func fetchAlbums() -> [Album] {
+    fileprivate func fetchSortedAlbums() -> [Album] {
```
a-szotyori: 👍 for clarity!

GiniVision/Classes/Core/GiniVisionUtils.swift
```diff
@@ -70,7 +70,9 @@ internal func UIFontPreferred(_ weight: FontWeight, andSize size: CGFloat) -> UI
 
 internal struct AnimationDuration {
     static var slow = 1.0
+    static var mediumSlow = 0.7
```
a-szotyori: Do we really need this granularity? Three speeds should be enough, I mean when do I use mediumSlow? Or should I use slow? Maybe mediumFast is better... no it's too slow I should use fast instead... Can be confusing to other devs.

GiniVision/Classes/Core/MultipageReviewTransitionAnimator.swift
```diff
@@ -9,7 +9,7 @@ import UIKit
 
 final class MultipageReviewTransitionAnimator: NSObject {
     
-    let animationDuration = AnimationDuration.medium
+    let animationDuration = AnimationDuration.mediumFast
```
a-szotyori: I am more a "well done" kind of guy, mediumFast ist just too rare for me. 🙃

GiniVision/Classes/Core/Gallery/GalleryCoordinator.swift
```diff
@@ -83,15 +89,26 @@ final class GalleryCoordinator: NSObject, Coordinator {
     // MARK: - Coordinator lifecycle
     
     func start() {
-        if let firstAlbum = galleryManager.albums.first {
-            let imagePicker = createImagePicker(with: firstAlbum)
-            galleryManager.startCachingImages(for: firstAlbum)
-            galleryNavigator.pushViewController(imagePicker, animated: false)
+        DispatchQueue.global().async {
```
a-szotyori: DispatchQueues, DispatchQueues everywhere.

GiniVision/Classes/Core/Gallery/GalleryCoordinator.swift
```diff
@@ -206,13 +226,38 @@ extension GalleryCoordinator: ImagePickerViewControllerDelegate {
...
+            }
+        }
+        let imageDocument = GiniImageDocument(data: data,
+                                              imageSource: .external,
+                                              imageImportMethod: .picker,
+                                              deviceOrientation: nil)
+        self.selectedImageDocuments[asset.identifier] = imageDocument
+        
+        if !self.giniConfiguration.multipageEnabled {
+            DispatchQueue.main.async {
```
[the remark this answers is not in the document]
kikettas: It is something that really displeases me, and I think the best solution as you suggest is to split into different methods.

GiniVision/Classes/Core/Gallery/GalleryCoordinator.swift
```diff
@@ -163,41 +182,42 @@ extension GalleryCoordinator: UINavigationControllerDelegate {
...
-                self.selectedImageDocuments[asset.identifier] = imageDocument
-                
-                if !self.giniConfiguration.multipageEnabled {
+                    self.addSelected(asset, withData: data)
+                } else {
                     DispatchQueue.main.async {
-                        self.openImages()
+                        viewController.addToDownloadingItems(index: index)
+                    }
+                    self.galleryManager.fetchImageData(from: asset, isRemote: true) { [weak self] data in
```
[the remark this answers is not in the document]
kikettas: image

So all the method can be simplified a little bit.

GiniVision/Classes/Core/Gallery/GalleryManager.swift
kikettas: Yes, make sense for this case. 👍

GiniVision/Classes/Core/Gallery/GalleryManager.swift
kikettas: 🖖

GiniVision/Classes/Core/GiniVisionUtils.swift
kikettas: Honestly when I wrote this I thought it was a good idea, thinking about different animation speeds. But for regular animations doesn't make sense. I think I will revert it to the previous values, with the recommended value Google guidelines.

GiniVision/Classes/Core/Gallery/GalleryCoordinator.swift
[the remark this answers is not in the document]
kikettas: This is something that really displeases me, I like your proposal of splitting them into methods and I think that it will fit perfectly with the likely new approach for concurrency that will come in Swift 5 ([This](https://gist.github.com/lattner/31ed37682ef1576b16bca1432ea9f782))

PR #192, opened 2018-04-09
Multipage alert dialog

kikettas:
The first time that the user uses the multipage, it will be shown a alert dialog in the camera screen pointing out how multipage works.

Final title and message are not defined yet.

### How to test
Run the Example app, and see that the dialog appears.

### Merging
Automatic


GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -288,7 +288,11 @@ extension GiniScreenAPICoordinator: CameraViewControllerDelegate {
     }
     
     func cameraDidAppear(_ viewController: CameraViewController) {
-        showOnboardingIfNeeded()
+        if shouldShowOnBoarding() {
```
a-szotyori: What do you think of showing it after the onboarding?

kikettas: I was thinking about it, but I haven't decided on the flow yet. We should take a look on this before the release, since there are different possibilities regarding toolTips/dialogs...

PR #194, opened 2018-04-11
Refactor document validation

kikettas:
To avoid raveled and unreadable code between the components of the application, it has been decoupled the validation from the camera screen, being done outside of it. Also the pickers has been also decoupled from the camera so it is now more clear. This eases the validation and flow between different components, being easier to use with the Component API.

### How to test
Run the Example app in single page mode, disable internet connection in the device and try to analyze a document.

### Merging
Automatic

### DIFF
--- a/Example/GiniVision/AppCoordinator.swift
+++ b/Example/GiniVision/AppCoordinator.swift
@@ -77,7 +77,7 @@ final class AppCoordinator: Coordinator {
         
         // 3. Validate document
         do {
-            try document?.validate()
+            try document?.validate(giniConfiguration: self.giniConfiguration)
             showOpenWithSwitchDialog(forDocuments: [document!])
         } catch {
             showExternalDocumentNotValidDialog()
--- a/Example/GiniVision/AppCoordinator.swift
+++ b/Example/GiniVision/AppCoordinator.swift
@@ -76,11 +76,14 @@ final class AppCoordinator: Coordinator {
         popToRootViewControllerIfNeeded()
         
         // 3. Validate document
-        do {
-            try document?.validate(giniConfiguration: self.giniConfiguration)
-            showOpenWithSwitchDialog(forDocuments: [document!])
-        } catch {
-            showExternalDocumentNotValidDialog()
+        if let document = document {
+            do {
+                try GiniVisionDocumentValidator.validate(document,
+                                                         withConfig: self.giniConfiguration)
+                showOpenWithSwitchDialog(forDocuments: [document])
+            } catch {
+                showExternalDocumentNotValidDialog()
+            }
         }
     }
     
--- a/Example/GiniVision/AppCoordinator.swift
+++ b/Example/GiniVision/AppCoordinator.swift
@@ -78,8 +78,8 @@ final class AppCoordinator: Coordinator {
         // 3. Validate document
         if let document = document {
             do {
-                try GiniVisionDocumentValidator.validate(document,
-                                                         withConfig: self.giniConfiguration)
+                try GiniVision.validate(document,
+                                        withConfig: self.giniConfiguration)
                 showOpenWithSwitchDialog(forDocuments: [document])
             } catch {
                 showExternalDocumentNotValidDialog()
--- a/Example/Tests/GINIVisionDocumentTests.swift
+++ b/Example/Tests/GINIVisionDocumentTests.swift
@@ -11,6 +11,7 @@ import XCTest
 
 class GINIVisionDocumentTests: XCTestCase {
     
+    let giniConfiguration: GiniConfiguration = GiniConfiguration()
     var filePickerManager: DocumentPickerCoordinator {
         return DocumentPickerCoordinator()
     }
@@ -18,9 +19,10 @@ class GINIVisionDocumentTests: XCTestCase {
     func testExcedeedMaxFileSize() {
         let higherThan10MBData = generateFakeData(megaBytes: 12)
         
-        let fakeDocument = GiniPDFDocument(data: higherThan10MBData)
+        let pdfDocument = GiniPDFDocument(data: higherThan10MBData)
         
-        XCTAssertThrowsError(try fakeDocument.validate(),
+        XCTAssertThrowsError(try GiniVisionDocumentValidator.validate(pdfDocument,
+                                                                      withConfig: giniConfiguration),
                              "Files with a size lower than 10MB should be valid") { error in
             XCTAssert(error as? DocumentValidationError == DocumentValidationError.exceededMaxFileSize,
                       "should indicate that max file size has been exceeded")
@@ -30,9 +32,10 @@ class GINIVisionDocumentTests: XCTestCase {
     func testNotExcedeedMaxFileSize() {
         let lowerThanOrEqualTo10MBData = generateFakeData(megaBytes: 10)
         
-        let fakeDocument = GiniPDFDocument(data: lowerThanOrEqualTo10MBData)
+        let pdfDocument = GiniPDFDocument(data: lowerThanOrEqualTo10MBData)
 
-        XCTAssertThrowsError(try fakeDocument.validate(),
+        XCTAssertThrowsError(try GiniVisionDocumentValidator.validate(pdfDocument,
+                                                                      withConfig: giniConfiguration),
                              "Files with a size greater than 10MB should not be valid") { error in
             XCTAssert(error as? DocumentValidationError != DocumentValidationError.exceededMaxFileSize,
                       "should indicate that max file size has been exceeded")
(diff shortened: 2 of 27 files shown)


Example/GiniVision/AppCoordinator.swift
```diff
@@ -76,11 +76,14 @@ final class AppCoordinator: Coordinator {
...
         
         // 3. Validate document
-        do {
-            try document?.validate()
-            showOpenWithSwitchDialog(forDocuments: [document!])
-        } catch {
-            showExternalDocumentNotValidDialog()
+        if let document = document {
+            do {
+                try GiniVisionDocumentValidator.validate(document,
```
a-szotyori: We have the `GiniVision` class. I would add the validate method to that one instead. It's easier for clients when they have a few classes for interacting with the library. Also it's easier to find methods with fewer classes to check.

GiniVision/Classes/Core/Camera.swift
```diff
@@ -177,7 +177,9 @@ extension Camera: AVCaptureMetadataOutputObjectsDelegate {
         if let metadataObj = metadataObjects.first as? AVMetadataMachineReadableCodeObject,
             metadataObj.type == AVMetadataObjectTypeQRCode {
             let qrDocument = GiniQRCodeDocument(scannedString: metadataObj.stringValue)
-            didDetectQR?(qrDocument) 
+            DispatchQueue.main.async { [weak self] in
```
a-szotyori: Is `captureOutput` now called on a secondary thread?

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -14,20 +14,13 @@ import AVFoundation
...
      - parameter documents: One or several documents either captured or imported in
      the `CameraViewController`. They can contain an error produced in the validation process.
-     - parameter completion: `DocumentPickerCompletion` block used to check if there is an issue with
-     the captured documents. The completion block also has an inner block that is executed once the
-     picker has been dismissed when there are no errors.
      */
     @objc func camera(_ viewController: CameraViewController,
-                      didCaptureDocuments documents: [GiniVisionDocument],
-                      validationHandler: DocumentValidationHandler?)
+                      didCapture document: GiniVisionDocument)
```
a-szotyori: You changed the public API meaning clients will have to do changes even without using multi-page. Can you instead deprecate old methods and still use them and make the new methods mandatory for multi-page?

GiniVision/Classes/Core/GiniVisionDocument.swift
```diff
@@ -126,11 +118,10 @@ extension GiniVisionDocument {
      - Throws: `DocumentValidationError.exceededMaxFileSize` is thrown if the document is not valid.
      
      */
-    public func validate() throws {
-        let document = self
+    public class func validate(_ document: GiniVisionDocument, withConfig giniConfiguration: GiniConfiguration) throws {
```
a-szotyori: The old signature should be deprecated and add a new method with new signature.

Example/Tests/GINIVisionDocumentTests.swift
```diff
@@ -11,16 +11,18 @@ import XCTest
...
     }
     
     func testExcedeedMaxFileSize() {
         let higherThan10MBData = generateFakeData(megaBytes: 12)
         
-        let fakeDocument = GiniPDFDocument(data: higherThan10MBData)
+        let pdfDocument = GiniPDFDocument(data: higherThan10MBData)
         
-        XCTAssertThrowsError(try fakeDocument.validate(),
+        XCTAssertThrowsError(try GiniVisionDocumentValidator.validate(pdfDocument,
```
nsobadjiev: I'm not really sure this is the right place for these tests. You moved the validation to `GiniVisionDocumentValidator`. So there's nothing about validation in the document class anymore. If the class doesn't have validation, then its tests shouldn't test the validation.

Just move these tests to the `GiniVisionDocumentValidatorTests` class.

Example/GiniVision/AppCoordinator.swift
kikettas: Makes sense. I will keep the class for the validation, but I will link the `GiniVision` class to it.

GiniVision/Classes/Core/GiniVisionDocument.swift
kikettas: True. I will revert it back and add the deprecation attribute.

GiniVision/Classes/Core/GiniScreenAPICoordinator+Analysis.swift
```diff
@@ -0,0 +1,105 @@
...
+    func createImageAnalysisNoResultsScreen() -> ImageAnalysisNoResultsViewController {
+        let imageAnalysisNoResultsViewController: ImageAnalysisNoResultsViewController
+        let isCameraViewControllerLoaded: Bool = {
+            guard let cameraViewController = cameraViewController else {
+                return false
+            }
+            return screenAPINavigationController.viewControllers.contains(cameraViewController)
+        }()
+        
+        if isCameraViewControllerLoaded {
```
nsobadjiev: Ideally, I'd like to avoid code like this. But I don't have ideas how yet.

GiniVision/Classes/Core/GiniScreenAPICoordinator+Camera.swift
```diff
@@ -0,0 +1,292 @@
...
+        
+        let navigationController = UINavigationController(rootViewController: vc)
+        navigationController.applyStyle(withConfiguration: giniConfiguration)
+        navigationController.modalPresentationStyle = .overCurrentContext
+        screenAPINavigationController.present(navigationController, animated: true, completion: nil)
+    }
+    
+    private func showMultipageNewFeatureDialog() {
+        let alertDialog = AlertDialogController(giniConfiguration: giniConfiguration,
+                                                title: "This is the title",
```
nsobadjiev: Are these just sample values? Will they be replaced by real ones? Preferably localized?

GiniVision/Classes/Core/GiniScreenAPICoordinator+Camera.swift
```diff
@@ -0,0 +1,292 @@
...
+    func showNextScreenAfterPicking(documents: [GiniVisionDocument]) {
+        if let firstDocument = documents.first, let documentsType = documents.type {
+            switch documentsType {
+            case .image:
+                if let imageDocuments = self.visionDocuments as? [GiniImageDocument],
+                    let lastDocument = imageDocuments.last {
+                    if self.giniConfiguration.multipageEnabled {
+                        if let imageDocuments = self.visionDocuments as? [GiniImageDocument],
+                            lastDocument.isImported {
+                            self.showMultipageReview(withImageDocuments: imageDocuments)
```
nsobadjiev: A lot of if cases here. Maybe it could have been expressed more clearly? Maybe a combination of comments and splitting the logic in several methods?

GiniVision/Classes/Core/GiniScreenAPICoordinator+Camera.swift
```diff
@@ -0,0 +1,292 @@
...
+                if let error = error as? FilePickerError {
+                    switch error {
+                    case .maxFilesPickedCountExceeded, .mixedDocumentsUnsupported:
+                        let imageDocuments = self.visionDocuments.flatMap { $0 as? GiniImageDocument }
+                        
+                        if imageDocuments.isNotEmpty {
+                            positiveAction = {
+                                coordinator.dismissCurrentPicker {
+                                    let imageDocuments = self.visionDocuments.flatMap { $0 as? GiniImageDocument }
+                                    self.showMultipageReview(withImageDocuments: imageDocuments)
```
nsobadjiev: Same thing here.

GiniVision/Classes/Core/Camera.swift
kikettas: And it was called before, but since I changed the logic behind `didDetectQR`, I prefer to execute it in the main thread here and not inside of the closure itself.

GiniVision/Classes/Core/GiniScreenAPICoordinator+Camera.swift
kikettas: In another branch I modified this if cases since it changes a little bit. Also, if you pay more attention to this, there is a redundant conditional unwrapping for the `imageDocuments`.
I won't change nothing here because it is already modify in another branch.

GiniVision/Classes/Core/GiniScreenAPICoordinator+Camera.swift
kikettas: The same here. This changes a little bit in another branch that will be merged after this one.

PR #204, opened 2018-04-24
Add new API multipage methods

kikettas:
Added new multipage document analysis methods, for partial and composite documents. 
The multipage review screen is loaded only once.

### How to test
To test this version, first you should get the development version of the API SDK and place it where the `Podfile` says.

### Merging
Automatic


GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -153,17 +186,16 @@ extension GiniScreenAPICoordinator {
...
     @objc func showAnalysisScreen() {
-        let documentToShow = visionDocuments[0]
-        if let didReview = visionDelegate?.didReview(document:withChanges:) {
-            didReview(documentToShow, changesOnReview)
-        } else if let didReview = visionDelegate?.didReview(_:withChanges:) {
-            didReview(documentToShow.data, changesOnReview)
+        if let didReviewDocuments = visionDelegate?.didReview(documents:) {
+            didReviewDocuments(visionDocuments)
+        } else if let didReview = visionDelegate?.didReview(document:withChanges:) {
+            didReview(visionDocuments[0], false)
```
nsobadjiev: I know we already discussed this in the past, but are you sure this is fine? Even if you're sure it will never be empty, maybe it still makes sense to have an assert to be extra explicit?

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -183,16 +215,25 @@ extension GiniScreenAPICoordinator: UINavigationControllerDelegate {
                               to toVC: UIViewController) -> UIViewControllerAnimatedTransitioning? {
         if fromVC == analysisViewController && operation == .pop {
             analysisViewController = nil
-            visionDelegate?.didCancelAnalysis?()
+            visionDelegate?.didCancelAnalysis()
         }
         if fromVC == reviewViewController && toVC == cameraViewController {
+            // This only happen when not using multipage
```
nsobadjiev: This CAN only happen...

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -183,16 +215,25 @@ extension GiniScreenAPICoordinator: UINavigationControllerDelegate {
...
             reviewViewController = nil
-            visionDelegate?.didCancelReview?()
-            visionDocuments.removeAll()
+            if let firstDocument = visionDocuments.first {
```
nsobadjiev: Here, you use `first` and not `[0]` like above. I think it's better.

GiniVision/Classes/Core/GiniVision.swift
```diff
@@ -20,15 +20,15 @@ import UIKit
...
+    func didCapture(document: GiniVisionDocument)
     
     /**
      Called when the user has taken an image.
      
      - parameter fileData: JPEG image data including meta information or PDF data
      */
-    @available(*, deprecated)
-    @objc optional func didCapture(_ imageData: Data)
+    @available(*, unavailable, renamed: "didCapture(document:)")
```
nsobadjiev: Breaking change?

GiniVision/Classes/Core/GiniVision.swift
```diff
@@ -38,14 +38,30 @@ import UIKit
...
      - parameter fileData:  JPEG image data including eventually updated meta information or PDF Data
      - parameter changes:   Indicates whether `imageData` was altered.
      */
-    @available(*, deprecated)
-    @objc optional func didReview(_ imageData: Data, withChanges changes: Bool)
+    @available(*, unavailable, renamed: "didReview(document:withChanges:)")
+    func didReview(_ imageData: Data, withChanges changes: Bool)
+    
+    /**
+     Called when the user is presented with the analsis screen. Use the `analysisDelegate`
```
nsobadjiev: <strike> analsis </strike> analysis

GiniVision/Classes/Core/ImageMetaInformationManager.swift
```diff
@@ -266,7 +266,7 @@ internal class ImageMetaInformationManager {
         return existingUUID ?? NSUUID().uuidString
     }
     
-    fileprivate func imageRotationDeltaDegrees() -> Int {
+    func imageRotationDeltaDegrees() -> Int {
```
nsobadjiev: Why would an external class need to know the rotation?

GiniVision/Classes/Core/MultipageReviewViewController.swift
```diff
@@ -226,6 +239,12 @@ extension MultipageReviewViewController {
                                          scrollPosition: .centeredHorizontally)
         self.collectionView(self.pagesCollection, didSelectItemAt: indexPath)
     }
+    
+    func reloadCollections() {
+        self.mainCollection.reloadData()
+        self.pagesCollection.reloadData()
+        self.selectItem(at: 0)
```
nsobadjiev: What if the collections are empty?

GiniVision/Classes/Core/GiniVision.swift
a-szotyori: https://bigidesignsblog.files.wordpress.com/2011/04/cartmananalprobesatalliteweb.jpg

Example/Podfile
```diff
@@ -5,6 +5,7 @@ platform :ios, '8.0'
 use_frameworks!
 
 target 'GiniVision_Example' do
+    pod 'Gini-iOS-SDK', :path => '../../gini-sdk-ios/'
```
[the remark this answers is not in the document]
kikettas: It is the only way to test this branch, at least now (If I can release an alpha version of the Gini API SDK I would change this)

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -32,18 +32,22 @@ internal final class GiniScreenAPICoordinator: NSObject, Coordinator {
     var cameraViewController: CameraViewController?
     var imageAnalysisNoResultsViewController: ImageAnalysisNoResultsViewController?
     var reviewViewController: ReviewViewController?
-    var multiPageReviewController: MultipageReviewViewController?
+    lazy var multiPageReviewViewController: MultipageReviewViewController = {
+        let imageDocuments = self.visionDocuments.flatMap { $0 as? GiniImageDocument }
```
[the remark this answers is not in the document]
kikettas: It is not possible to have a mix different documents right now. Worst case scenario would be to have an empty array, but won't happen.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -32,18 +32,22 @@ internal final class GiniScreenAPICoordinator: NSObject, Coordinator {
     var cameraViewController: CameraViewController?
     var imageAnalysisNoResultsViewController: ImageAnalysisNoResultsViewController?
     var reviewViewController: ReviewViewController?
-    var multiPageReviewController: MultipageReviewViewController?
+    lazy var multiPageReviewViewController: MultipageReviewViewController = {
```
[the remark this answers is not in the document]
kikettas: Yes. It also concerned me but I think it makes totally sense since it is probably the screen with more interaction. Also I think that in terms of performance is much worse to create the view controller every time.

GiniVision/Classes/Core/GiniVision.swift
kikettas: It is a breaking change, but given the feedback I got regarding this delegate, the best thing is to remove the optional methods when it's possible and make the old ones `unavailable`.

If you deprecate a protocol method, Xcode doesn't complain about it. And it could take more time to realize that the method is missing than just using the Xcode assistant to fix it.

GiniVision/Classes/Core/MultipageReviewViewController.swift
kikettas: This is already fixed in another PR.

GiniVision/Classes/Core/ImageMetaInformationManager.swift
kikettas: The `GiniImageDocument` class needs to know the rotation angle in order to be used by other classes.

GiniVision/Classes/Core/Extensions/Collection.swift
```diff
@@ -35,5 +35,21 @@ extension Collection where Iterator.Element == GiniVisionDocument {
     var type: GiniVisionDocumentType? {
         return containsDifferentTypes ? nil : first?.type
     }
+}
+
+extension Array where Iterator.Element == GiniVisionDocument {
     
+    mutating func remove(_ document: GiniVisionDocument) {
+        if let documentIndex = (self.index { $0.id == document.id }) {
```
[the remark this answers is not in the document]
kikettas: I had several problems making a protocol `Hashable` and `Equatable`, due to some issue with Obj-C interoperability. That is why I cannot do what you suggest 😢

GiniVision/Classes/Core/Extensions/Collection.swift
```diff
@@ -35,5 +35,21 @@ extension Collection where Iterator.Element == GiniVisionDocument {
...
+            remove(at: documentIndex)
+        }
+    }
+    
+    mutating func updateOrAdd(_ document: GiniVisionDocument) {
+        if let documentIndex = (self.index { $0.id == document.id }) {
```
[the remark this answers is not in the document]
kikettas: Answered above!

GiniVision/Classes/Core/GiniPDFDocument.swift
```diff
@@ -32,6 +33,7 @@ final public class GiniPDFDocument: NSObject, GiniVisionDocument {
     init(data: Data) {
         self.data = data
         self.isReviewable = false
+        self.id = UUID().uuidString
```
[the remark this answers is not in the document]
kikettas: Extensions in Swift cannot have stored properties (there is a workaround but it would be more confusing in this case). This is the main downside of not being able to use generics 😞.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
kikettas: The thing is that I haven't decided yet if I want to make that method also unavailable and keep only the one that handle one or several documents at the same time. Of course looking for the first element is completely wrong. I will change it for this PR, but I will probably delete it in a future PR.

Example/Podfile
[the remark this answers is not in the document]
nsobadjiev: You can always specify a commit, tag or branch to take the SDK from.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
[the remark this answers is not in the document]
nsobadjiev: But then why have that code altogether? It conveys the wrong idea (that having a mixture is possible)

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -32,18 +32,22 @@ internal final class GiniScreenAPICoordinator: NSObject, Coordinator {
...
+        let imageDocuments = self.visionDocuments.flatMap { $0 as? GiniImageDocument }
+        let multiPageReviewViewController =
+            self.createMultipageReviewScreenContainer(withImageDocuments: imageDocuments)
+        _ = multiPageReviewViewController.view
```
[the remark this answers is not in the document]
nsobadjiev: It's premature optimisation. Do you know how much time you'd safe?
My guess would be that it's not big enough to justify sacrificing code quality.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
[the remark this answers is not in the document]
nsobadjiev: Then you have to be really careful to clear the state after you're done with the controller. Because it would be really bad if you use the same one afterward but it still has some data from the previous screen.

And again, it seems like premature optimisation.

GiniVision/Classes/Core/MultipageReviewViewController.swift
nsobadjiev: Fixed in another PR? What does that mean? Why isn't it fixed here?

GiniVision/Classes/Networking/DocumentServiceProtocol.swift
```diff
@@ -0,0 +1,187 @@
...
+    func fetchExtractions(for documents: [PartialDocumentInfo],
+                          completion: @escaping AnalysisCompletion) {
+        let partialDocumentsInfo = documents.map { $0.toDictionary() }
+        analysisCancellationToken = BFCancellationTokenSource()
+        
+        giniSDK
+            .documentTaskManager
+            .createCompositeDocument(withPartialDocumentsInfo: partialDocumentsInfo,
+                                     fileName: "",
+                                     docType: "",
```
nsobadjiev: Why are the file name and doc type always empty?

GiniVision/Classes/Networking/DocumentServiceProtocol.swift
```diff
@@ -0,0 +1,187 @@
...
+            .continueOnSuccessWith(block: { _ in
+                return self.giniSDK
+                .documentTaskManager?
+                    .update(self.compositeDocument,
+                            updatedExtractions: updatedExtractions,
+                            cancellationToken: nil)
+            })
+            .continueWith(block: { (task: BFTask?) in
+                guard let extractions = task?.result as? NSMutableDictionary else {
+                    print("Error sending feedback for document with id: ",
```
nsobadjiev: Don't you always put a ❌ in these logs?

GiniVision/Classes/Networking/Extensions/GiniScreenAPICoordinator+Networking.swift
```diff
@@ -37,15 +37,15 @@ extension GiniScreenAPICoordinator {
         }
     }
     
-    var apiService: APIServiceProtocol? {
+    var documentService: DocumentServiceProtocol? {
         get {
-            return objc_getAssociatedObject(self, &AssociatedKey.apiService) as? APIServiceProtocol
+            return objc_getAssociatedObject(self, &AssociatedKey.documentService) as? DocumentServiceProtocol
```
nsobadjiev: I don't understand why we need the runtime here.

GiniVision/Classes/Networking/Extensions/GiniScreenAPICoordinator+Networking.swift
```diff
@@ -128,52 +118,68 @@ extension GiniScreenAPICoordinator: GiniVisionDelegate {
...
+        
+        if giniConfiguration.multipageEnabled {
+            uploadDocumentCompletionHandler = { result in
+                switch result {
+                case .success:
+                    break
+                case .failure(let error):
+                    print("Partial document creation error: ", error)
+                }
+                // TODO: Update upload status in MultipageReviewViewController
```
nsobadjiev: Can't you do that now?

GiniVision/Classes/Networking/SinglePageDocumentsService.swift
```diff
@@ -0,0 +1,71 @@
...
+    func update(parameters: [String: Any], for document: GiniVisionDocument) {
+        partialDocumentInfo?.updateAdditionalParameters(with: parameters)
+    }
+    
+    func upload(document: GiniVisionDocument,
+                completion: UploadDocumentCompletion?) {
+        partialDocumentInfo = PartialDocumentInfo()
+        
+        createDocument(from: document,
+                       fileName: "fileName") { result in
```
nsobadjiev: File name is "fileName"...
I really don't like it - it seems to me that the file name property is useless.

GiniVision/Classes/Networking/SinglePageDocumentsService.swift
```diff
@@ -0,0 +1,71 @@
...
+            switch result {
+            case .success(let document):
+                self.partialDocumentInfo?.documentUrl = document.links.document
+                
+                if let handler = self.pendingAnalysisHandler {
+                    self.startAnalysis(completion: handler)
+                }
+            case .failure(let error):
+                print("Partial document creation error: ", error)
```
nsobadjiev: Again, no ❌?

GiniVision/Classes/Networking/DocumentServiceProtocol.swift
kikettas: For testing purposes. But I can add a default fileName,

kikettas: Actually in the next PR I added a "FileName" because it was failing when testing.

GiniVision/Classes/Networking/Extensions/GiniScreenAPICoordinator+Networking.swift
kikettas: Because it is the only way to add stored properties to protocols.

GiniVision/Classes/Networking/Extensions/GiniScreenAPICoordinator+Networking.swift
kikettas: It is exactly the next PR

GiniVision/Classes/Core/GiniPDFDocument.swift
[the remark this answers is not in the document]
kikettas: I think that UUID is clear enough in this case right? There is no need to abstract it now.

GiniVision/Classes/Core/MultipageReviewViewController.swift
kikettas: There is a PR depending on this one.

Example/Podfile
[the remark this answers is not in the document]
kikettas: True! I forgot about the `:branch` attribute.

https://media.giphy.com/media/3oFzmkkwfOGlzZ0gxi/giphy.gif

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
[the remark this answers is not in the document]
kikettas: Actually for this case, doesn't make sense to load the view right away. We can wait until the view controller presentation.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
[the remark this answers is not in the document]
kikettas: You will never have documents from previous session. When using this (screen API), you can only go back to multipage screen if you had a problem in the analysis. Which makes sense, since you want to have the same documents. Once the analysis has been completed, the Screen API should be deallocated.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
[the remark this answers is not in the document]
kikettas: Now a mixture is not possible, but it is quite likely to have it in the future.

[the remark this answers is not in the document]
nsobadjiev: Then, in the future you can change it. I'd still consider adding an assertion to make sure it doesn't happen accidentally

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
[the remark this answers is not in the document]
nsobadjiev: So you will remove it?

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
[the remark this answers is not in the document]
nsobadjiev: Well, if you like it this way keep it.
From my experience it brings more problems than solutions, but it's not a big deal.

GiniVision/Classes/Networking/Extensions/GiniScreenAPICoordinator+Networking.swift
nsobadjiev: Is it absolutely necessary to have it in an extension then?

Example/Podfile
```diff
@@ -5,7 +5,7 @@ platform :ios, '8.0'
 use_frameworks!
 
 target 'GiniVision_Example' do
-    pod 'Gini-iOS-SDK', :path => '../../gini-sdk-ios/'
+    pod 'Gini-iOS-SDK', :git => 'git@github.com:gini/gini-sdk-ios.git', :branch => 'develop'
```
nsobadjiev: Cool, just make sure to add a version, not a branch before merging `multipage`

GiniVision/Classes/Networking/DocumentServiceProtocol.swift
nsobadjiev: Hehe, nice!

GiniVision/Classes/Networking/Extensions/GiniScreenAPICoordinator+Networking.swift
kikettas: Yes. It is the optional Networking package, so some classes wouldn't be available in the Core package.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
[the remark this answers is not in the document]
kikettas: Yes, sorry.

Example/Podfile
kikettas: Added an issue in Github so I won't forget about it before release.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -33,10 +33,11 @@ internal final class GiniScreenAPICoordinator: NSObject, Coordinator {
     var imageAnalysisNoResultsViewController: ImageAnalysisNoResultsViewController?
     var reviewViewController: ReviewViewController?
     lazy var multiPageReviewViewController: MultipageReviewViewController = {
-        let imageDocuments = self.visionDocuments.flatMap { $0 as? GiniImageDocument }
+        guard let imageDocuments = self.visionDocuments as? [GiniImageDocument] else {
+            fatalError("The MultipageReviewViewController can only handle image documents.")
```
nsobadjiev: Fatal error will crash even in production. This might be a good thing in many cases, but I think not in this one.
You want to catch these errors while developing, but in production you have a good backup plan. You just filter the non-images and continue.

So I'd return the old code that was filtering the collection but also assert if there is a non-image.

PR #206, opened 2018-05-03
Uploading state in pages collection

kikettas:
Added new state view in pages collection, being possible now to see if a document has been uploaded or not.
Also added scrollview inside the main collection to make the image zoomable.

### How to test
Run the example app, take a picture and see if the state of the upload in the multipage review view controller.

### Merging
Automatic


Example/Tests/Extensions.swift
```diff
@@ -30,4 +30,18 @@ internal extension XCTestCase {
...
     }
+    
+    func loadValidatedImageDocument(withName name: String) -> ValidatedDocument {
+        let path = Bundle.main.url(forResource: name, withExtension: "jpg")
+        let data = try? Data(contentsOf: path!)
+        let builder = GiniVisionDocumentBuilder(data: data, documentSource: .external)
+        return ValidatedDocument(value: (builder.build() as? GiniImageDocument)!)
+    }
+    
+    func loadValidatedPDFDocument(withName name: String) -> ValidatedDocument {
```
nsobadjiev: This method is really similar to `loadValidatedImageDocument`. Merge them? Not necessarily have one method, but rather merge the contents of the method.

Example/Tests/PageStateViewTests.swift
```diff
@@ -0,0 +1,41 @@
...
+    var statusView: PageStateView!
+    
+    override func setUp() {
+        super.setUp()
+        statusView = PageStateView(frame: .zero)
+    }
+    
+    func testLoadingState() {
+        statusView.update(to: .loading)
+        XCTAssertNil(statusView.icon.image, "icon image should be nil when it is loading")
```
nsobadjiev: I'm not sure that this tests. Just setting the state doesn't mean the image is shown or even set. And as far as I saw, it's just an empty image view at that point. So I'd argue that this assertion is not really useful.

Same thing applies to the following two methods.

GiniVision/Classes/Core/GiniScreenAPICoordinator+Camera.swift
```diff
@@ -134,10 +144,10 @@ extension GiniScreenAPICoordinator: CameraViewControllerDelegate {
     
     private func showMultipageNewFeatureDialog() {
         let alertDialog = AlertDialogController(giniConfiguration: giniConfiguration,
-                                                title: "This is the title",
-                                                message: "This is the subtitle",
+                                                title: "Multi-Page Rechnungsanalyse",
+                                                message: "Jetzt ist es möglich, eine Rechnung mit mehreren Seiten zu analysieren",
                                                 image: UIImageNamedPreferred(named: "multipageIcon"),
-                                                buttonTitle: "Let's scan!",
+                                                buttonTitle: "Zur Kamera",
```
nsobadjiev: All strings here are not localized and/or configurable.

GiniVision/Classes/Core/PageStateView.swift
```diff
@@ -0,0 +1,75 @@
...
+//
+//  Created by Enrique del Pozo Gómez on 4/13/18.
+//
+
+import Foundation
+
+final class PageStateView: UIView {
+    
+    enum State {
+        case success, failure, loading
```
nsobadjiev: I'm just being picky, but the naming is not consistent. Success and failure are nouns and then you have "loading", which is not. And from a spoken language perspective, the question is "What's the state of this page?" an answer would be "It failed / It's loading / it succeeded". I think the general convention would be to have:

- completed
- failed
- loading

GiniVision/Classes/Core/ValidatedDocument.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  ValidatedDocument.swift
+//  GiniVision
+//
+//  Created by Enrique del Pozo Gómez on 4/12/18.
+//
+
+import Foundation
+
+public struct ValidatedDocument {
```
nsobadjiev: What's the point here? A Validated document is a wrapper around the Gini VisionDocument that also has information about the validation and the state?

It's not really easily understandable and maybe even confusing.

GiniVision/Classes/Networking/DocumentServiceProtocol.swift
```diff
@@ -91,21 +91,20 @@ extension DocumentServiceProtocol {
             })
     }
     
-    func fetchExtractions(for documents: [PartialDocumentInfo],
+    func fetchExtractions(for documents: [GINIPartialDocumentInfo],
```
nsobadjiev: A class prefix? Why?

GiniVision/Classes/Networking/MultipageDocumentsService.swift
```diff
@@ -29,41 +32,47 @@ final class MultipageDocumentsService: DocumentServiceProtocol {
...
+
     }
     
     func upload(document: GiniVisionDocument,
                 completion: UploadDocumentCompletion?) {
         let cancellationTokenSource = BFCancellationTokenSource()
         let token = cancellationTokenSource.token
-        self.partialDocuments[document.id] = PartialDocumentInfo()
+        self.partialDocuments[document.id] =
+            PartialDocumentInfo(info: (GINIPartialDocumentInfo(documentId: nil, rotationDelta: 0)),
```
nsobadjiev: Ok, so there's `PartialDocumentInfo ` and `GINIPartialDocumentInfo `. That's confusing!

GiniVision/Classes/Core/GiniScreenAPICoordinator+Camera.swift
kikettas: This is part of the Dialog that is removed in a future PR as I told you.

GiniVision/Classes/Core/PageStateView.swift
kikettas: You are right, I was biased with the naming and didn't pay attention to it. 👍

Example/Tests/PageStateViewTests.swift
kikettas: Makes sense for the first one but yeah, for the next two tests we should check the image itself

GiniVision/Classes/Core/ValidatedDocument.swift
kikettas: I couldn't come up with a better naming. Do you have any idea?

nsobadjiev: I would argue that it is confusing to have it there.
Did you create it after I commented that a document shouldn't have an error because it's a breach of capsulation?

GiniVision/Classes/Networking/MultipageDocumentsService.swift
nsobadjiev: Why can't you just use one of them?

kikettas: Because I don't want to let `GINIPartialDocumentInfo` object know about the order, since it is not part of it.

GiniVision/Classes/Core/ValidatedDocument.swift
kikettas: I think that the problem here is just the name, since it is not a `GiniVisionDocument` but a container for the document and associated value (an error in this case)

PR #208, opened 2018-05-07
UI Changes in Camera Screen

kikettas:
After the user testing, some UI changes were required in the camera screen to improve the UX, such as subtitles for icons, stack indicator, shadow for preview image and going directly to multipage screen after first image captured.

### How to test

### Merging
Automatic

### Issues
#198 #200 #201


nsobadjiev:
> [...] some UI changes were required in the camera screen.

What kind of changes?

Example/Tests/CapturedImagesStackViewTests.swift
```diff
@@ -0,0 +1,50 @@
...
+    
+    override func setUp() {
+        super.setUp()
+        capturedImagesStackView = CapturedImagesStackView(frame: .zero)
+    }
+    
+    func testCaptureStackWhenNoImages() {
+        capturedImagesStackView.updateStackStatus(to: .empty)
+        
+        XCTAssertTrue(capturedImagesStackView.isHidden,
```
nsobadjiev: I'm not sure if a view should hide itself (from a responsibility perspective). As a standalone component it might be a strange behaviour.

On the other hand, it's quite convenient here.

Just a comment...

Example/Tests/CapturedImagesStackViewTests.swift
```diff
@@ -0,0 +1,50 @@
...
+    
+    func testCaptureStackWhenTwoImageCaptured() {
+        capturedImagesStackView.updateStackStatus(to: .filled(count: 2,
+                                                              lastImage: loadImage(withName: "invoice.jpg")!))
+        XCTAssertFalse(capturedImagesStackView.isHidden,
+                       "capturedImagesStackView should not be hidden when it is filled")
+        XCTAssertFalse(capturedImagesStackView.thumbnailStackBackgroundView.isHidden,
+                       "thumbnailStackBackgroundView should not be hidden when there are 2 images")
+
+    }
```
nsobadjiev: What I'm missing are test about setting the actual images.
Maybe also that they are shown in the right sequence (the last one on top).

Example/Tests/GINICameraViewControllerTests.swift
```diff
@@ -53,32 +53,5 @@ final class CameraViewControllerTests: XCTestCase {
...
-    func testReviewButtonBackgroundAfter1ImageWasCaptured() {
-        _ = cameraViewController.view
-
-        cameraViewController.cameraDidCapture(imageData: imageData, error: nil)
-        XCTAssertTrue(cameraViewController.multipageReviewBackgroundView.isHidden,
-                      "multipageReviewBackgroundView should be hidden after capture the first picture")
-        
-    }
-    
-    func testReviewButtonBackgroundAfter2ImagesWereCaptured() {
```
nsobadjiev: Are these tests really already invalid? It might be interesting to test the integration between getting the event that an image has been captured and seeing the change reflected in the review stack view.

GiniVision/Classes/Core/CapturedImagesStackView.swift
```diff
@@ -0,0 +1,178 @@
...
+
+final class CapturedImagesStackView: UIView {
+    
+    enum State {
+        case filled(count: Int, lastImage: UIImage), empty
+    }
+    
+    let thumbnailSize: CGSize = {
+        if UIDevice.current.isIpad {
+            return CGSize(width: 40, height: 60)
```
nsobadjiev: Can't you get away with not hardcoding this? Just infer it from the view size.

GiniVision/Classes/Core/CapturedImagesStackView.swift
```diff
@@ -0,0 +1,178 @@
...
+    
+    fileprivate lazy var capturedImagesStackSubtitleLabel: UILabel = {
+        let label = UILabel()
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.text = NSLocalizedString("ginivision.camera.capturedImagesStackLabel",
+                                       bundle: Bundle(for: GiniVision.self),
+                                       comment: "label shown below images stack")
+        label.textAlignment = .center
+        label.textColor = .white
+        label.font = label.font.withSize(12)
```
nsobadjiev: Didn't you create an enum for common font sizes at one point?

GiniVision/Classes/Core/CapturedImagesStackView.swift
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// MARK: - Public methods
+
+extension CapturedImagesStackView {
+    func absoluteThumbnailFrame(from view: UIView) -> CGRect {
+        return convert(thumbnailButton.frame, to: view)
+    }
+    
+    func updateStackStatus(to status: State) {
```
nsobadjiev: I'm not a fan of this interface. I have to create a State variable to change the stack view? It seems like an overkill and it's a bit confusing. 

I'd prefer something similar to the next method, `addImageToStack`.
The most common use case is to add or remove an image (maybe also reorder or replace).

GiniVision/Classes/Core/CapturedImagesStackView.swift
```diff
@@ -0,0 +1,178 @@
...
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("Use init(frame:) initializer instead")
+    }
+
+}
+
+// MARK: - Public methods
+
+extension CapturedImagesStackView {
+    func absoluteThumbnailFrame(from view: UIView) -> CGRect {
```
nsobadjiev: Initially, I didn't like this method. It is confusing and breaks encapsulation. I understand that there's no better way to create the transition animation, but I think it at least warrants a comment.

Example/Tests/GINICameraViewControllerTests.swift
kikettas: That is why I removed them, they should not be part of the unit test since it is the integration between two components.

GiniVision/Classes/Core/CapturedImagesStackView.swift
kikettas: The customization is not part of this PR since it has not been defined yet. What you mean is a `customFont` property in the `GiniConfiguration`.
This will probably be customizable, but not decided yet.

GiniVision/Classes/Core/CapturedImagesStackView.swift
kikettas: You are right, also does not make sense to have different ways of updating it.

GiniVision/Classes/Core/CapturedImagesStackView.swift
kikettas: Don't understand this. Why does it break encapsulation? `CapturedImagesStackView` is providing a way of getting its position from other view's perspective.

Example/Tests/CapturedImagesStackViewTests.swift
kikettas: 👍

GiniVision/Classes/Core/CapturedImagesStackView.swift
kikettas: No, because I want them to be that size.

GiniVision/Classes/Core/CapturedImagesStackView.swift
nsobadjiev: It's not really "absolute". It's not relative to the screen, but to a superview.

I just thought that it was a strange method. I'd put a comment saying that it's related to animations and you can get the frame of the thumbnail so you can use it for animating. But just a remark from my side, not an issue.

Example/Tests/GINICameraViewControllerTests.swift
kikettas: We don't have integration tests yet, IMO is worst to mess up here because we will forget to remove it in the future.

Example/Tests/CapturedImagesStackViewTests.swift
```diff
@@ -19,31 +19,38 @@ final class CapturedImagesStackViewTests: XCTestCase {
...
-                                                              lastImage: loadImage(withName: "invoice.jpg")!))
+        let images = [loadImage(withName: "invoice.jpg")!]
+        
+        capturedImagesStackView.replaceStackImages(with: images)
         
         XCTAssertFalse(capturedImagesStackView.isHidden,
                       "capturedImagesStackView should not be hidden when it is filled")
         XCTAssertTrue(capturedImagesStackView.thumbnailStackBackgroundView.isHidden,
                       "thumbnailStackBackgroundView should be hidden when there is only 1 image")
+        XCTAssertEqual(capturedImagesStackView.thumbnailButton.image(for: .normal), images[0],
```
nsobadjiev: I disagree a little bit here. You don't have to test the image in all methods - it's redundant. And now you have three assertions.

I would argue that you have 3 tests:
* having one image -> you test that the stack view is not hidden
* having two images -> you test that the background view is not hidden
* having at least one image -> you test that the image view shows the thumbnail of the top most image.

GiniVision/Classes/Core/CapturedImagesStackView.swift
kikettas: 👍 Added [issue](https://github.com/gini/gini-vision-lib-ios/issues/209) to tackle this in a future PR.

GiniVision/Classes/Core/CapturedImagesStackView.swift
kikettas: Maybe I could make it variable depending on container height (for iPhone) or width (for iPad), but due to time constraints and the complexity that this would add to the class, I will keep it as it is now.

GiniVision/Classes/Core/CapturedImagesStackView.swift
nsobadjiev: I'm sorry, but I'm still not satisfied. The comment is too generic - I see what the method does, but I don't understand why it exists. I'd say, even though is not super loosely coupled, just write that it's about being able to make a transition from the thumbnail to a new screen.

Also, as I pointed out, you are not returning an absolute frame, but rather a frame relative to another view. So the name is misleading. I'd use `frameRelative:to:` or something.

PR #210, opened 2018-05-08
UI changes in multipage review

kikettas:
After the user testing, some UI changes were required in the multipage screen to improve the UX, such as fixed pages collection, add images button and drag and drop tip.

### How to test

### Merging
Automatic

### Issues
#203 #202


GiniVision/Assets/Localizable.strings
```diff
@@ -98,5 +98,6 @@
 "ginivision.imagepicker.openbutton" = "Öffnen";
 
 "ginivision.multipagereview.reorderContainerTooltipMessage" = "Die Seiten müssen in der richtigen Reihenfolge sein. Hier können Sie sie sortieren";
-
-
+"ginivision.multipagereview.addButtonLabel" = "Fügen Sie dieser Rechnung eine neue Seite hinzu";
+"ginivision.multipagereview.dragAndDropTip" = "Drag & Drop zum Sortieren";
+"ginivision.multipagereview.title" = "%d von %d";
```
nsobadjiev: I'm usually not a fan of sprintf formatting in strings files. And here you might just get away with localizing the string in the middle.

The only problem is that there might be languages where the format is different and the string is not in the middle. However, even this format might not be sufficient for all languages.

What I'm saying is that I'm not sure about this, but I don't know how to proceed. We are unlikely to encounter such languages any time soon.

GiniVision/Classes/Core/Multipage Review/MultipageReviewPagesCollectionFooter.swift
```diff
@@ -0,0 +1,153 @@
...
+//
+//  Created by Enrique del Pozo Gómez on 4/26/18.
+//
+
+import Foundation
+
+final class MultipageReviewPagesCollectionFooter: UICollectionReusableView {
+    
+    static let identifier = "MultipageReviewPagesCollectionFooterIdentifier"
+    var didTapAddButton: (() -> Void)?
```
nsobadjiev: Again, you are mixing delegates and callbacks in the same project. You are missing a little consistency. But it's ok, I guess.

GiniVision/Classes/Core/Multipage Review/MultipageReviewPagesCollectionFooter.swift
```diff
@@ -0,0 +1,153 @@
...
+        let width = MultipageReviewPagesCollectionFooter.contentSize.width +
+            padding.left +
+            padding.right
+        
+        return CGSize(width: width, height: height)
+    }
+    
+    class func padding(in collectionView: UICollectionView? = nil) -> UIEdgeInsets {
+        var rightPadding: CGFloat = 10
+        if let collection =  collectionView {
```
nsobadjiev: Double spaces!

GiniVision/Classes/Core/Multipage Review/MultipageReviewPagesCollectionFooter.swift
```diff
@@ -0,0 +1,153 @@
...
+            rightPadding = ((collection.frame.width -
+                MultipageReviewPagesCollectionCell.size.width) / 2)
+        }
+        
+        return UIEdgeInsets(top: 10, left: 10, bottom: 10, right: rightPadding)
```
nsobadjiev: All these tens displease me. I'd use `defaultPadding = 10` and have a `rightPadding` that's calculated.

GiniVision/Classes/Core/Multipage Review/MultipageReviewViewController.swift
```diff
@@ -72,6 +72,19 @@ public final class MultipageReviewViewController: UIViewController {
         return view
     }()
     
+    lazy var pagesCollectionBottomTipLabel: UITextView = {
+        let textView = UITextView()
+        textView.translatesAutoresizingMaskIntoConstraints = false
+        textView.text = NSLocalizedString("ginivision.multipagereview.dragAndDropTip",
+                                       bundle: Bundle(for: GiniVision.self),
+                                       comment: "drag and drop tip shown below pages collection")
+        textView.font = textView.font?.withSize(11)
```
nsobadjiev: That's what I was talking about in the last PR. Now the font is 11, not 12. I believe a consistent size would have been a little better.

kikettas: As I said in the previous PR, I will extract this values to a class in a different PR.

GiniVision/Assets/Localizable.strings
kikettas: Localizing strings in iOS is a mess. Maybe when I create the enum for the strings an images, I can handle this.

PR #211, opened 2018-05-08
UI changes in onboarding and no results screens.

kikettas:
After the user testing, some UI changes were required in the onboarding screen to improve the UX, such as the background, a new suggestion and no dialog for multipage feature.

### How to test

### Merging
Automatic

### Issues
#197 #196


Example/Tests/GINIImageNoResultsViewControllerTests.swift
```diff
@@ -35,6 +35,7 @@ final class GINIImageNoResultsViewControllerTests: XCTestCase {
     
     override func setUp() {
         _ = viewController.view
+        viewController.captureSuggestions = items
```
nsobadjiev: I have a problem with these tests in general. 

First of all, I don't see a benefit in using the real values used by the app also at runtime.

Secondly, and more important, I don't think it makes sense to have a test for each suggestion. It duplicates tests and the only benefit is that theoretically someone might exchange the indexes. I'd recommend having only one test that tests one of the cells. Maybe not the first one, just pick one.

GiniVision/Classes/Core/GiniConfiguration.swift
```diff
@@ -423,6 +423,13 @@ import UIKit
         NSLocalizedStringPreferred("ginivision.onboarding.fourthPage",
                                    comment: "Text on the fourth page of the onboarding screen")
     
+    /**
+     Sets the text on the fifth onboarding page.
+     */
+    public var onboardingFifthPageText =
```
nsobadjiev: It's a bit strange to have five method for modifying each element in an array. It'd opt for having an array property where all values are set at the same time.

I know it's already part of the API, but we can consider deprecating these methods and adding the one described above.

P.S. Same applies for the suggestions.

PR #212, opened 2018-05-08
Update to Swift 4.1

kikettas:
Updated project and code to Swift 4.1. 
To make the example app for Obj-C compile, some classes has been removed (they would have been removed in a future PR either way).

### How to test
Run the example app and check that everything works as it did before.

### Merging
Automatic

### Issues
#162


GiniVision/Classes/Core/Camera.swift
```diff
@@ -99,7 +99,7 @@ internal class Camera: NSObject {
             self.stillImageOutput?
                 .captureStillImageAsynchronously(from: connection) { (buffer: CMSampleBuffer?, error: Error?) in
                     guard error != nil else {
-                        completion(AVCaptureStillImageOutput.jpegStillImageNSDataRepresentation(buffer), nil)
+                        completion(AVCaptureStillImageOutput.jpegStillImageNSDataRepresentation(buffer!), nil)
```
nsobadjiev: I'd use the guard that's already here to also unwrap the buffer.
I also think that the iOS SDK has many APIs where not having an error object doesn't mean everything went smoothly. I'm not sure this is the case here, but I think it would have been cleaner if you used the guard.

PR #214, opened 2018-05-09
Add review controller delegate

kikettas:
In order to be consistent with the rest of screens, it has been added a delegate for the `ReviewViewController`.

### How to test
Nothing to test

### Merging
Automatic


nsobadjiev:
Are you sure there's nothing to test? You could add tests to see if the delegate methods are invoked.
Maybe even that the callbacks still work (they might be deprecated, but they still should work).

kikettas:
@nsobadjiev You are right, maybe it is good if the delegate and blocks are tested when the rotate is done. 
Added them! 👍

PR #217, opened 2018-05-15
Update ObjC example with last changes

kikettas:
Update the ObjC example app with the last version of the GVL, adding minor fixes in the GVL regarding ObjC-Swift interoperability.

### How to test
Open the app and test that multipage works.

### Merging
Automatic


Example/Example ObjC/ScreenAPIViewController.m
```diff
@@ -48,134 +45,48 @@ - (IBAction)easyLaunchGiniVision:(id)sender {
...
+    giniConfiguration.openWithEnabled = YES;
+    giniConfiguration.qrCodeScanningEnabled = YES;
     
-    // 2. Create the Gini Vision Library view controller, set a delegate object and pass in the configuration object
-    UIViewController *vc = [GiniVision viewControllerWithDelegate:self withConfiguration:giniConfiguration importedDocument: NULL];
+    NSDictionary<NSString*, NSString*> *credentials = [[[CredentialsManager alloc] init] getCredentials];
     
+    GiniClient *client = [[GiniClient alloc] initWithClientId:credentials[@"client_id"]
+                                                 clientSecret:credentials[@"client_password"]
+                                            clientEmailDomain:credentials[@"client_domain"]];
```
nsobadjiev: Didn't you make these key names constants for the Swift example? They look sad hardcoded like this.

Example/Example ObjC/ScreenAPIViewController.m
```diff
@@ -48,134 +45,48 @@ - (IBAction)easyLaunchGiniVision:(id)sender {
...
+- (void)giniVision:(NSArray<id<GiniVisionDocument>> *)documents
+analysisDidFinishWithResults:(NSDictionary<NSString *,GINIExtraction *> *)results
+      sendFeedback:(void (^)(NSDictionary<NSString *,GINIExtraction *> * _Nonnull))sendFeedback {
+    _result = results;
+    [self presentResultsWithSendFeedbackBlock:sendFeedback];
 }
 
-- (void)presentResults {
+- (void)presentResultsWithSendFeedbackBlock:(SendFeedbackBlock)sendFeedbackBlock {
     NSArray *payFive = @[@"paymentReference", @"iban", @"bic", @"amountToPay", @"paymentRecipient"];
```
nsobadjiev: Do we still need to check for the pay5? I vaguely remember that we decided to work even if some of them are missing. Moreover, the BIC (and even the recipient for that matter) is not mandatory to make a transfer.

Example/Example ObjC/ScreenAPIViewController.m
```diff
@@ -189,7 +100,7 @@ - (void)presentResults {
     if (hasPayFive) {
         ResultTableViewController *vc = [storyboard instantiateViewControllerWithIdentifier:@"resultScreen"];
```
nsobadjiev: At some point we should consider putting all storyboard names and identifiers in enums. Do you want to create an issue?

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -222,9 +217,15 @@ extension GiniScreenAPICoordinator: UINavigationControllerDelegate {
                               from fromVC: UIViewController,
                               to toVC: UIViewController) -> UIViewControllerAnimatedTransitioning? {
         if fromVC == analysisViewController && operation == .pop {
+            // Going directly from the analysis to the camera means that
+            // the document is not an image and should be removed
+            if toVC == cameraViewController {
```
a-szotyori: We will need to move scattered network related calls to dedicated classes (extensions) in the future. In the Android version we have logic scattered around, too. If we are not careful it might byte us in the future.

Can you create a ticket for this? For the Android version we created a [similar one](https://github.com/gini/gini-vision-lib-android/issues/190).

Example/Example ObjC/ScreenAPIViewController.m
kikettas: It is merely an example, they may choose to take only the IBAN. But I guess that adding a comment can clarify it.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
nsobadjiev: I'd like to second that comment. I was really hoping to avoid such code but I keep seeing it.

kikettas: But there is not networking here. Which makes me think that the name for that elements is still wrong. I created [a ticket](https://github.com/gini/gini-vision-lib-ios/issues/219) in order to find a better name.

PR #221, opened 2018-05-23
UI and Qr code scanning fixes.

kikettas:
Some minor issues were fixed:
- Qr codes are not processed properly
- Gallery shows the correct button when deselecting images
- Captured images stack thumbnail now shows the correct aspect
- View controllers now have a black background
- _Weiter_ button state is updated when documents are added/deleted/modified
- Validation indicator now is shown smoothly and with a short delay.

### How to test
Open the app and test that multipage works.

### Merging
Automatic

### Issues
#220


GiniVision/Classes/Core/GiniScreenAPICoordinator+Review.swift
```diff
@@ -12,7 +12,7 @@ import Foundation
 extension GiniScreenAPICoordinator: ReviewViewControllerDelegate {
     
     func review(_ viewController: ReviewViewController, didReview document: GiniVisionDocument) {
-        self.updateValueInDocuments(for: document)
+        updateValue(for: document)
```
nsobadjiev: A name such as `updateValue` doesn't give me much information about what the method does.

GiniVision/Classes/Core/GiniScreenAPICoordinator+Review.swift
```diff
@@ -94,13 +94,12 @@ extension GiniScreenAPICoordinator: MultipageReviewViewControllerDelegate {
         }
     }
     
-    func refreshMultipageReview(with documentRequests: [DocumentRequest]) {
+    func refreshMultipageReviewNextButton(with documentRequests: [DocumentRequest]) {
```
nsobadjiev: I would have been ok with the old name. It doesn't have to specify everything it does. It's just a method for updating the UI. The specifics can be encapsulated.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -136,33 +136,42 @@ final class GiniScreenAPICoordinator: NSObject, Coordinator {
 
 extension GiniScreenAPICoordinator {
     func addToDocuments(new documentRequests: [DocumentRequest]) {
+        if self.documentRequests.type == .qrcode || self.documentRequests.type == .pdf {
```
nsobadjiev: Ideally I'd avoid such edge cases. If you can, I'd prefer it if you could find a way to process everything in the same way.

GiniVision/Classes/Core/GiniScreenAPICoordinator+Review.swift
kikettas: Before it made sense, because it did different things. But know it is just the next button refreshing.

GiniVision/Classes/Core/GiniScreenAPICoordinator+Review.swift
kikettas: Since I will rename the `DocumentRequest`, it will make more sense to call it `updateDocument`

PR #222, opened 2018-05-24
Create logger

kikettas:
Replace `print` calls with `Logger` calls.  

### How to test
Open the app, and use it to see that the Logger is working.

### Merging
Automatic

### Issues
#167

### DIFF
--- a/GiniVision/Classes/Core/Logger.swift
+++ b/GiniVision/Classes/Core/Logger.swift
@@ -1,0 +1,46 @@
+//
+//  Logger.swift
+//  GiniVision
+//
+//  Created by Enrique del Pozo Gómez on 5/14/18.
+//
+
+import Foundation
+
+final class Logger {
+    
+    enum Event {
+        case error
+        case success
+        case warning
+        case custom(emoji: String)
+        
+        var value: String {
+            switch self {
+            case .error: return "❌"
+            case .success: return "✅"
+            case .warning: return "⚠️"
+            case .custom(let emoji): return emoji
+            }
+        }
+    }
+    
+    class func debug(message: String,
+                     event: Event,
+                     giniConfig: GiniConfiguration = .shared) {
+        
+        if giniConfig.debugModeOn {
+            print("[ GiniVision ](\(formattedString(from: Date()))): \(event.value) \(message)")
+        }
+    }
+    
+    class func formattedString(from date: Date) -> String {
+        let dateFormat = "dd-MM-yy hh:mm:ss"
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = dateFormat
+        dateFormatter.locale = Locale.current
+        dateFormatter.timeZone = TimeZone.current
+        
+        return dateFormatter.string(from: date)
+    }
+}
--- a/GiniVision/Classes/Core/Logger.swift
+++ b/GiniVision/Classes/Core/Logger.swift
@@ -32,7 +32,12 @@ final class Logger {
                    giniConfig: GiniConfiguration = .shared) {
         
         if giniConfig.debugModeOn {
-            NSLog("[ GiniVision ] \(event.value) \(message)")
+            let message = "[ GiniVision ] \(event.value) \(message)"
+            if let loggerBlock = giniConfig.customLog {
+                loggerBlock(message)
+            } else {
+                NSLog(message)
+            }
         }
     }
 }
--- a/GiniVision/Classes/Core/Logger.swift
+++ b/GiniVision/Classes/Core/Logger.swift
@@ -16,6 +16,7 @@ enum LogEvent: String {
 func Log(message: String,
          event: LogEvent,
          giniConfig: GiniConfiguration = .shared) {
+
     Log(message: message, event: event.rawValue, giniConfig: giniConfig)
 }
 
@@ -24,11 +25,29 @@ func Log(message: String,
          giniConfig: GiniConfiguration = .shared) {
     
     if giniConfig.debugModeOn {
+        giniConfig.logger.log(message: message, event: event)
+    }
+}
+
+@objc public protocol GiniLogger: class {
+    
+    /**
+     Logs a message
+     
+     - parameter message: Message printed out
+     - parameter event: String which represents the printed message. i.e: ⚠️
+     
+     */
+    func log(message: String, event: String)
+}
+
+final class DefaultLogger: GiniLogger {
+    
+    func log(message: String, event: String) {
         let message = "[ GiniVision ] \(event) \(message)"
-        if let loggerBlock = giniConfig.customLog {
-            loggerBlock(message)
-        } else {
-            NSLog(message)
+        if ProcessInfo.processInfo.environment["OS_ACTIVITY_MODE"] == "disable" {
+            print(message)
         }
+        NSLog(message)
     }
 }
--- a/GiniVision/Classes/Core/Logger.swift
+++ b/GiniVision/Classes/Core/Logger.swift
@@ -25,7 +25,7 @@ func Log(message: String,
          giniConfig: GiniConfiguration = .shared) {
     
     if giniConfig.debugModeOn {
-        giniConfig.logger.log(message: message, event: event)
+        giniConfig.logger.log(message: "\(event) \(message)")
     }
 }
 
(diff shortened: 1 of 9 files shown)


GiniVision/Classes/Core/Logger.swift
```diff
@@ -0,0 +1,46 @@
...
+            switch self {
+            case .error: return "❌"
+            case .success: return "✅"
+            case .warning: return "⚠️"
+            case .custom(let emoji): return emoji
+            }
+        }
+    }
+    
+    class func debug(message: String,
```
nsobadjiev: I don't see a point in having a debug method if there is no other log level. I'm not saying that we need log levels (but I'm also not saying that we shouldn't have them).

The point is that if you don't want logging levels, just have a generic `log` method.

GiniVision/Classes/Core/Logger.swift
```diff
@@ -0,0 +1,46 @@
...
+                     giniConfig: GiniConfiguration = .shared) {
+        
+        if giniConfig.debugModeOn {
+            print("[ GiniVision ](\(formattedString(from: Date()))): \(event.value) \(message)")
+        }
+    }
+    
+    class func formattedString(from date: Date) -> String {
+        let dateFormat = "dd-MM-yy hh:mm:ss"
+        let dateFormatter = DateFormatter()
```
nsobadjiev: As far as I remember, it wasn't recommended to create Date Formatters all the time because they are "expensive" objects. You should consider having a static instance for the class.

GiniVision/Classes/Networking/DocumentServiceProtocol.swift
```diff
@@ -58,8 +58,8 @@ extension DocumentServiceProtocol {
                                                                                cancellationToken: cancellationToken)
             }).continueWith(block: { task in
                 if let createdDocument = task.result as? GINIDocument {
-                    print("📄 Created document with id:", createdDocument.documentId ?? "",
-                          "for vision document:", document.id)
+                    Logger.debug(message: "Created document with id: \(createdDocument.documentId ?? "") " +
+                        "for vision document \(document.id)", event: .custom(emoji:"📄"))
```
nsobadjiev: I think your Event enum breaks a little bit here. Instead of just putting the emoji itself, you need to have a whole statement - `.custom(emoji:"📄")`. It's kind of verbose. Maybe you can find a way to shorten it and have only the character as parameter?

GiniVision/Classes/Core/Logger.swift
```diff
@@ -0,0 +1,46 @@
...
+    
+    enum Event {
+        case error
+        case success
+        case warning
+        case custom(emoji: String)
+        
+        var value: String {
+            switch self {
+            case .error: return "❌"
```
nsobadjiev: Have you tried extracting the logs from a real device (from iTunes or Xcode)? Can you verify that these emojis show? It's a real use case that someone gets logs from a device (from a crash reporter for instance) and tries to make sense of them.

GiniVision/Classes/Core/Logger.swift
kikettas: Yes, makes sense since we are not going to introduce more log levels.

GiniVision/Classes/Core/Logger.swift
kikettas: Didn't know about its bad performance 👍

kikettas: I removed it since it is not needed with `NSLog` (it already prints the date)

GiniVision/Classes/Core/Logger.swift
kikettas: They are shown if I use `NSLog` (the whole message, including emojis).

kikettas:
@nsobadjiev These are the logs that I consider important, because I don't what to show a lot of logs in the console (they might be annoying and pointless)

GiniVision/Classes/Core/Logger.swift
nsobadjiev: `NSLog` might be another thing. It logs directly to `stderr`, but it's possible that a partner has a custom logging solution that writes somewhere else. And they don't want to redirect `stderr`. So it might be useful to be able to inject a logging function. But maybe I'm overthinking it.

GiniVision/Classes/Networking/DocumentServiceProtocol.swift
```diff
@@ -179,13 +179,13 @@ extension DocumentServiceProtocol {
...
-                    Logger.debug(message: message, event: .error)
+                    Logger.log(message: message, event: .error)
                     
                     return nil
                 }
                 
-                Logger.debug(message: "Feedback sent with \(updatedExtractions.count) extractions",
-                    event: .custom(emoji:"🚀"))
+                Logger.log(message: "Feedback sent with \(updatedExtractions.count) extractions",
+                    event: .custom("🚀"))
```
nsobadjiev: I'm still a little skeptical about this. It's a little shorter, but still a bit verbose. It's way longer than just having the emoji inline in the string. And actually, do you really need custom ones?

GiniVision/Classes/Core/Logger.swift
kikettas: Ummm, could make sense although it is not very likely. I added a custom log block to the `GiniConfiguration` so they can customize it.

GiniVision/Classes/Networking/DocumentServiceProtocol.swift
```diff
@@ -179,13 +179,13 @@ extension DocumentServiceProtocol {
                 if let error = task?.error {
                     let id = self.compositeDocument?.documentId ?? ""
                     let message = "Error sending feedback for document with id: \(id) error: \(error)"
-                    Logger.debug(message: message, event: .error)
+                    Logger.log(message: message, event: .error)
```
nsobadjiev: What about having a function for it. So we don't always have to write `Logger.log`, which can be annoying. We could replace it with `Log(...)` or something.

GiniVision/Classes/Networking/DocumentServiceProtocol.swift
kikettas: The idea is to represent the event in a graphical way, but since there is not way to check if it is an emoji or just a string maybe it makes sense to pass just a `String`.

GiniVision/Classes/Networking/DocumentServiceProtocol.swift
kikettas: I like the idea! 💯

GiniVision/Classes/Core/Logger.swift
```diff
@@ -7,40 +7,28 @@
...
+
+func Log(message: String,
+         event: LogEvent,
+         giniConfig: GiniConfiguration = .shared) {
+    Log(message: message, event: event.rawValue, giniConfig: giniConfig)
+}
+
+func Log(message: String,
+         event: String,
+         giniConfig: GiniConfiguration = .shared) {
```
nsobadjiev: No one would pass a configuration object to a logger method, I think. Can you make it a property in the class?

kikettas: I don't understand, what's that class?

nsobadjiev: I wanted to say the Logger class. But it's no longer there.
I'm not sure why you removed it though. You could have kept it and had a `Log` method that uses it.

kikettas:
It is failing randomly due to a problem with a file. Not sure if it is because of the CI or the repo.

GiniVision/Classes/Core/Logger.swift
```diff
@@ -0,0 +1,53 @@
...
+@objc public protocol GiniLogger: class {
+    
+    /**
+     Logs a message
+     
+     - parameter message: Message printed out
+     - parameter event: String which represents the printed message. i.e: ⚠️
+     
+     */
+    func log(message: String, event: String)
```
nsobadjiev: I'd avoid having partners implement a method with two parameters. If they have a custom logging solution, they probably have a method that takes one parameters. To also take an event type, they will also have to write an additional method to comply. Whereas, if it was only one, they'd likely just need to formally "say" that their logger complies to `GiniLogger `

kikettas: Then I will concat both string internally.

PR #224, opened 2018-05-29
Add errorview to multipage screen

kikettas:
Add error view in each page of the multipage screen.

### How to test
Open the app, select an invalid image and see the error in the multipage screen.

### Merging
Automatic

### DIFF
--- a/GiniVision/Classes/Core/Multipage Review/MultipageReviewMainCollectionCell.swift
+++ b/GiniVision/Classes/Core/Multipage Review/MultipageReviewMainCollectionCell.swift
@@ -62,7 +62,7 @@ final class MultipageReviewMainCollectionCell: UICollectionViewCell {
         }
     }
     
-    func fill(with documentRequest: DocumentRequest, errorAction: @escaping () -> Void) {
+    func fill(with documentRequest: DocumentRequest, errorAction: @escaping (ErrorAction) -> Void) {
         documentImage.image = documentRequest.document.previewImage
         
         let currentNoticeView = subviews
@@ -76,11 +76,14 @@ final class MultipageReviewMainCollectionCell: UICollectionViewCell {
         }
     }
     
-    func showErrorView(with currentNoticeView: NoticeView?, error: Error, errorAction: @escaping () -> Void) {
+    func showErrorView(with currentNoticeView: NoticeView?,
+                       error: Error,
+                       errorAction: @escaping (ErrorAction) -> Void) {
+        
         let newNoticeView = noticeView(with: error, errorAction: errorAction)
         
         if let currentNoticeView = currentNoticeView {
-            currentNoticeView.hide(completion: {
+            currentNoticeView.hide(false, completion: {
                 self.addSubview(newNoticeView)
                 newNoticeView.show(false)
             })
@@ -91,14 +94,17 @@ final class MultipageReviewMainCollectionCell: UICollectionViewCell {
     }
     
     func noticeView(with error: Error,
-                    errorAction: @escaping () -> Void) -> NoticeView {
+                    errorAction: @escaping (ErrorAction) -> Void) -> NoticeView {
         let buttonTitle: String
+        let action: ErrorAction
         
         switch error {
-        case is DocumentValidationError:
-            buttonTitle = "Retake"
-        default:
+        case is AnalysisError:
             buttonTitle = "Retry"
+            action = .retry
+        default:
+            buttonTitle = "Retake"
+            action = .retake
         }
         
         let message: String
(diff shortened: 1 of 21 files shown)


GiniVision/Classes/Core/Multipage Review/MultipageReviewPagesCollectionCell.swift
```diff
@@ -220,4 +216,16 @@ final class MultipageReviewPagesCollectionCell: UICollectionViewCell {
...
     }
+    
+    class func size(in collection: UICollectionView) -> CGSize {
+        let collectionInset = UIEdgeInsets(top: 16, left: 0, bottom: 16, right: 0)
+        let height = collection.frame.height -
+            collectionInset.top -
+            collectionInset.bottom +
+            shadowHeight +
+        shadowRadius
+        let width = height * 11 / 20
```
nsobadjiev: `let width = height * aspectRatio`? Wouldn't it be clearer?

GiniVision/Classes/Core/Multipage Review/MultipageReviewViewController.swift
```diff
@@ -415,29 +435,34 @@ extension MultipageReviewViewController {
...
+            self.pagesCollection.deleteItems(at: [indexPath])
+        }, completion: { [weak self] _ in
+            guard let `self` = self else { return }
+            if self.documentRequests.count > 0 {
+                if indexPath.row != self.documentRequests.count {
+                    var indexes = IndexPath.indexesBetween(indexPath,
+                                                           and: IndexPath(row: self.documentRequests.count,
+                                                                          section: 0))
+                    indexes.append(indexPath)
+                    self.pagesCollection.reloadItems(at: indexes)
```
nsobadjiev: Why do you reload? Isn't the whole point on the batch updates to avoid it and be able to animated?
And isn't it dangerous to reload items in a completion handler?

GiniVision/Classes/Core/Multipage Review/MultipageReviewViewController.swift
```diff
@@ -415,29 +435,34 @@ extension MultipageReviewViewController {
...
+        documentRequests.remove(at: indexPath.row)
+        mainCollection.deleteItems(at: [indexPath])
+        
+        pagesCollection.performBatchUpdates({
+            self.pagesCollection.deleteItems(at: [indexPath])
+        }, completion: { [weak self] _ in
+            guard let `self` = self else { return }
+            if self.documentRequests.count > 0 {
+                if indexPath.row != self.documentRequests.count {
+                    var indexes = IndexPath.indexesBetween(indexPath,
```
nsobadjiev: This whole completion handler is really difficult to understand. It does so many things and has different conditions that are difficult to follow.

GiniVision/Classes/Core/Multipage Review/MultipageReviewViewController.swift
```diff
@@ -415,29 +435,34 @@ extension MultipageReviewViewController {
...
+                                                           and: IndexPath(row: self.documentRequests.count,
+                                                                          section: 0))
+                    indexes.append(indexPath)
+                    self.pagesCollection.reloadItems(at: indexes)
+                }
+                
+                self.selectItem(at: min(indexPath.row, self.documentRequests.count - 1))
+            }
+            self.delegate?.multipageReview(self, didDelete: documentToDelete)
+            completion?()
```
nsobadjiev: A completion handler, that calls a delegate method and calls another completion handler? It's almost too much!

GiniVision/Classes/Core/Multipage Review/MultipageReviewMainCollectionCell.swift
```diff
@@ -62,6 +75,52 @@ final class MultipageReviewMainCollectionCell: UICollectionViewCell {
         }
     }
     
+    func fill(with documentRequest: DocumentRequest, errorAction: @escaping (NoticeActionType) -> Void) {
```
nsobadjiev: The name is misleading. It sounds like a purely visual property, but in reality it populates the cell.

GiniVision/Classes/Core/Multipage Review/MultipageReviewPagesCollectionCell.swift
kikettas: It is. 👍

GiniVision/Classes/Core/Multipage Review/MultipageReviewViewController.swift
kikettas: The page indicator in each cell should be refreshed and this cannot be done because you cannot delete and refresh the same `IndexPath`. (If I delete `1`, `2` should be `1` __after__ the deletion). 
It is not dangerous, the `performBatchUpdates` method just executes all the animations in the same `loop`, but you can execute more animations once the animation block finishes.

GiniVision/Classes/Core/Multipage Review/MultipageReviewViewController.swift
kikettas: I changed the order of the animation, so I can get rid of the completion handler and make it more clear.

nsobadjiev:
The build fails... again!

PR #227, opened 2018-06-01
Fixed open with and QRcode/PDF document upload

kikettas:
When using open with, the documents were not being uploaded. Also there were problems with PDF and QR code document upload, since the analysis might start when the document wasn't uploaded.

### How to test
Open _Files_ app and open a file from there. Also try to analyze a QR code and check that you get results.

### Merging
Automatic

### DIFF
--- a/Example/Example Swift/Info.plist
+++ b/Example/Example Swift/Info.plist
@@ -49,7 +49,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>0.8.2</string>
+	<string>0.9.0</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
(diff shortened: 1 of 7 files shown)


Example/Example Swift/Info.plist
```diff
@@ -49,7 +49,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>0.8.2</string>
+	<string>0.9.0</string>
```
nsobadjiev: What's the reasoning behind the sample app versioning? Isn't it about time it got a version that's not 0.X?

kikettas:
@nsobadjiev Yes, that was a mistake. I should have created a branch only for the beta release.

Example/Example Swift/Info.plist
kikettas: Just to identify that there is a new version for the beta, but it will be 1.0.0 when the new GVL version is released.

nsobadjiev: Just don't forget it.

PR #230, opened 2018-06-01
Merge master into develop

kikettas:
No changes, just update the `develop` branch.


nsobadjiev:
Why would you merge from master to develop? It should be the other way around? You shouldn't be developing in master!

PR #232, opened 2018-06-04
Add missing customizations

kikettas:
Some customizations have been added to new screens like Multipage review screen. Also has been fixed the value of the view controllers background.

### How to test
Change the value of these custom properties in the `GiniConfiguration` and check that they change in the view.

### Merging
Automatic

### DIFF
--- a/Example/Tests/GINIMultipageReviewControllerTests.swift
+++ b/Example/Tests/GINIMultipageReviewControllerTests.swift
@@ -141,13 +141,46 @@ final class GINIMultipageReviewControllerTests: XCTestCase {
                        "Fifth toolbar item should be the deleteButton")
     }
     
-    func testToolBarAndPagesCollectionContainerColors() {
-        XCTAssertEqual(multipageReviewViewController.toolBar.barTintColor, multipageReviewViewController.pagesCollectionContainer.backgroundColor,
-                       "toolbar and pages collection container background colors should match")
+    func testToolBarTintColor() {
+        let giniConfiguration = GiniConfiguration()
+        giniConfiguration.multipagePagesContainerAndToolBarColor = .black
+        let multipageReviewViewController = MultipageReviewViewController(documentRequests: [],
+                                                                          giniConfiguration: giniConfiguration)
+        
+        XCTAssertEqual(multipageReviewViewController.toolBar.barTintColor,
+                       giniConfiguration.multipagePagesContainerAndToolBarColor,
+                       "toolbar tint color should match the one specified in the configuration")
+    }
+    
+    func testPagesContainerBackgroundColor() {
+        let giniConfiguration = GiniConfiguration()
+        giniConfiguration.multipagePagesContainerAndToolBarColor = .black
+        let multipageReviewViewController = MultipageReviewViewController(documentRequests: [],
+                                                                          giniConfiguration: giniConfiguration)
+        
+        XCTAssertEqual(multipageReviewViewController.pagesCollectionContainer.backgroundColor,
+                       giniConfiguration.multipagePagesContainerAndToolBarColor,
+                       "pages container background color should match the one specified in the gini configuration")
+    }
+    
+    func testToolbarItemsColor() {
+        let giniConfiguration = GiniConfiguration()
+        giniConfiguration.multipageToolbarItemsColor = .black
+        let multipageReviewViewController = MultipageReviewViewController(documentRequests: [],
+                                                                          giniConfiguration: giniConfiguration)
+        
+        _ = multipageReviewViewController.view
+        XCTAssertEqual((multipageReviewViewController.deleteButton.customView as? UIButton)?.tintColor,
+                       giniConfiguration.multipageToolbarItemsColor,
+                       "delete button tint color should match the one specified in the gini configuration")
+        XCTAssertEqual((multipageReviewViewController.rotateButton.customView as? UIButton)?.tintColor,
+                       giniConfiguration.multipageToolbarItemsColor,
+                       "rotate button tint color should match the one specified in the gini configuration")
     }
     
     func testDatasourceOnDelete() {
-        let vc = MultipageReviewViewController(documentRequests: imageDocumentRequests, giniConfiguration: giniConfiguration)
+        let vc = MultipageReviewViewController(documentRequests: imageDocumentRequests,
+                                               giniConfiguration: giniConfiguration)
         _ = vc.view
         vc.view.setNeedsLayout()
         vc.view.layoutIfNeeded()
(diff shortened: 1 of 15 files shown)


Example/Tests/GINIMultipageReviewControllerTests.swift
```diff
@@ -194,6 +227,25 @@ final class GINIMultipageReviewControllerTests: XCTestCase {
...
+        _ = viewController.view
+        
+        let cell = viewController
+            .collectionView(multipageReviewViewController.pagesCollection,
+                            cellForItemAt: IndexPath(row: 0, section: 0)) as? MultipageReviewPagesCollectionCell
+        
+        XCTAssertEqual(cell?.pageIndicatorLabel.textColor, giniConfiguration.multipagePageIndicatorColor,
+                       "page cell indicator color should match the one specified in the configuration")
+        
+        XCTAssertEqual(cell?.bottomContainer.backgroundColor, giniConfiguration.multipagePageBackgroundColor,
```
nsobadjiev: These are actually two different tests, aren't they? One for the text color and one for the background.

Example/Tests/ImagePickerCollectionViewCellTests.swift
```diff
@@ -0,0 +1,36 @@
...
+        XCTAssertEqual(cell.checkCircleBackground.backgroundColor,
+                       giniConfiguration.galleryPickerItemSelectedBackgroundCheckColor)
+    }
+    
+    func testCellCheckIndicatorBackgroundOnDeselected() {
+        let giniConfiguration = GiniConfiguration()
+        
+        cell.changeCheckCircle(to: false, giniConfiguration: giniConfiguration)
+        XCTAssertEqual(cell.checkImage.alpha, 0)
+        XCTAssertEqual(cell.checkCircleBackground.backgroundColor,
```
nsobadjiev: Did you quit on adding assertion comments for this file :D?

GiniVision/Classes/Core/AnalysisViewController.swift
```diff
@@ -135,7 +135,7 @@ import UIKit
         super.loadView()
         imageView.image = document.previewImage
         edgesForExtendedLayout = []
-        view.backgroundColor = .black
+        view.backgroundColor = GiniConfiguration.shared.backgroundColor
```
nsobadjiev: I'm a bit confused. There are some view controllers where the configuration is injected in the constructor and here, the shared one is used. Is there a reason for this? If not, I'd say that it would have been better to be consistent and inject it here too.

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -279,7 +279,7 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> Void
     public override func loadView() {
         super.loadView()
         edgesForExtendedLayout = []
-        view.backgroundColor = .black
+        view.backgroundColor = GiniConfiguration.shared.backgroundColor
```
nsobadjiev: Here, it gets even worse. The configuration **is** injected, but here you use the shared instance, This can lead to subtle bugs.

GiniVision/Classes/Core/GiniNavigationViewController.swift
```diff
@@ -25,9 +25,7 @@ internal class GiniNavigationViewController: UINavigationController {
         navigationBar.tintColor = GiniConfiguration.shared.navigationBarItemTintColor
         var attributes = navigationBar.titleTextAttributes ?? [NSAttributedStringKey: Any]()
         attributes[NSAttributedStringKey.foregroundColor] = GiniConfiguration.shared.navigationBarTitleColor
-        attributes[NSAttributedStringKey.font] = GiniConfiguration.shared.customFont.isEnabled ?
-            GiniConfiguration.shared.customFont.light.withSize(16) :
-            GiniConfiguration.shared.navigationBarTitleFont
+        attributes[NSAttributedStringKey.font] = GiniConfiguration.shared.navigationBarTitleFont
```
nsobadjiev: Here you're using the shared instance again.

GiniVision/Classes/Core/GiniConfiguration.swift
```diff
@@ -584,6 +594,28 @@ import UIKit
      */
     @objc public var reviewTextBottomColor = UIColor.white
     
+    // MARK: Multipage options
+    
+    /**
+     Sets the color of the pages container and toolbar
```
nsobadjiev: I know you're working on an updated integration guide / documentation. It might make sense to expand on these comments a little more. They don't provide more information than the variable names. Which view is that container and where and how does the toolbar show.

Example/Tests/GINIMultipageReviewControllerTests.swift
```diff
@@ -141,13 +141,46 @@ final class GINIMultipageReviewControllerTests: XCTestCase {
                        "Fifth toolbar item should be the deleteButton")
     }
     
-    func testToolBarAndPagesCollectionContainerColors() {
-        XCTAssertEqual(multipageReviewViewController.toolBar.barTintColor, multipageReviewViewController.pagesCollectionContainer.backgroundColor,
-                       "toolbar and pages collection container background colors should match")
+    func testToolBarTintColor() {
```
nsobadjiev: Didn't we drop the `GINI` file prefix? Why is it still in this one?

GiniVision/Classes/Core/AnalysisViewController.swift
kikettas: I saw this in the AnalysisViewController, but I was thinking on creating a different PR for this. Maybe it makes sense to add it in this one 👍

GiniVision/Classes/Core/GiniNavigationViewController.swift
kikettas: This class is not used and removed in a different PR.

Example/Tests/GINIMultipageReviewControllerTests.swift
kikettas: Not yet, my idea is to do it once we merge `multipage` into develop.

Example/Tests/GINIMultipageReviewControllerTests.swift
kikettas: I always have the same doubt here, either create a different test for each one or use the same for both. My concern is the performance, since unit tests should always be fast.
What do you think?

Example/Tests/ImagePickerCollectionViewCellTests.swift
kikettas: 🤦‍♂️

GiniVision/Classes/Core/GiniConfiguration.swift
kikettas: That is a good idea, but would take a lot of time. I added a lot of comments in the next PR, but I can add more in another PR. I created the issue for that.

Example/Tests/GINIMultipageReviewControllerTests.swift
nsobadjiev: I think you should measure how long this test takes. Because, as always, I advise against premature optimisation. These tests are creating views so it's possible that they take a bit of time, but I suspect they are really fast to the point where optimisation doesn't make sense.

That being said, there are many other test classes where you initialize a full view controller for every test, so either way it's happening everywhere.

Even if it turns out that these tests are slow, you can fix that by reusing cells between tests. Although you will have to be careful to reset the cell at the end of a test.

In general, I think performance will not be a big issue here.

Issue #223, opened 2018-05-28
Fix duplicated upload for QR code documents


kikettas:
Also fix issue with QR codes in Component API

PR #240, opened 2018-06-06
Fixed QR code and PDF documents analysis

kikettas:
When a user tried to analyze a PDF/QRCode document, it wasn't analyzed because the analysis wasn't triggered after the document was uploaded.
Also when scanning a QR code, this wasn't processed properly, being uploaded twice when it was validated (tap on the popup button triggered the upload again).

### How to test
Run the Example app, scan a QR code and check that you obtain results afterwards.

### Merging
Automatic

### Issue
#223

### DIFF
--- a/Example/Example Swift/ComponentAPICoordinator.swift
+++ b/Example/Example Swift/ComponentAPICoordinator.swift
@@ -37,7 +37,7 @@ final class ComponentAPICoordinator: NSObject, Coordinator {
         navBarViewController.navigationBar.barTintColor = self.giniColor
         navBarViewController.navigationBar.tintColor = .white
         navBarViewController.view.backgroundColor = .black
-
+        
         return navBarViewController
     }()
     
@@ -74,7 +74,7 @@ final class ComponentAPICoordinator: NSObject, Coordinator {
     fileprivate(set) var reviewScreen: ReviewViewController?
     fileprivate(set) lazy var documentPickerCoordinator =
         DocumentPickerCoordinator(giniConfiguration: giniConfiguration)
-
+    
     init(documentRequests: [DocumentRequest],
          configuration: GiniConfiguration,
          client: GiniClient) {
@@ -213,13 +213,13 @@ extension ComponentAPICoordinator {
         
         push(viewController: vc, removingViewControllerOfType: AnalysisViewController.self)
         analysisScreen = nil
-
+        
     }
     
     fileprivate func showNextScreenAfterPicking() {
         if let documentsType = documentRequests.type {
             switch documentsType {
-            case .image:
+            case .image:                
                 if giniConfiguration.multipageEnabled {
                     refreshMultipageReview(with: self.documentRequests)
                     showMultipageReviewScreen()
@@ -354,7 +354,7 @@ extension ComponentAPICoordinator {
             closeComponentAPI()
             return
         }
-
+        
         navigationController.popToRootViewController(animated: true)
     }
 }
@@ -409,15 +409,25 @@ extension ComponentAPICoordinator: CameraViewControllerDelegate {
         validate([document]) { result in
             switch result {
             case .success(let documentRequests):
-                self.documentRequests.append(contentsOf: documentRequests)
-                self.upload(documentRequests: documentRequests)
-                
-                if self.giniConfiguration.multipageEnabled, self.documentRequests.count > 1 {
-                    if let imageDocument = document as? GiniImageDocument {
-                        viewController.animateToControlsView(imageDocument: imageDocument)
+                switch document {
+                case let qrCodeDocument as GiniQRCodeDocument:
+                    viewController.showPopup(forQRDetected: qrCodeDocument) {
+                        self.documentRequests.removeAll()
+                        self.documentRequests.append(contentsOf: documentRequests)
+                        self.upload(documentRequests: documentRequests)
+                        self.showNextScreenAfterPicking()
                     }
-                } else {
-                    self.showNextScreenAfterPicking()
+                case let imageDocument as GiniImageDocument:
+                    self.documentRequests.append(contentsOf: documentRequests)
+                    self.upload(documentRequests: documentRequests)
+                    
+                    if self.documentRequests.count > 1 {
+                        viewController.animateToControlsView(imageDocument: imageDocument)
+                    } else {
+                        self.showNextScreenAfterPicking()
+                    }
+                    
+                default: break
                 }
             case .failure(let error):
                 if let error = error as? FilePickerError, error == .maxFilesPickedCountExceeded {
@@ -480,13 +490,13 @@ extension ComponentAPICoordinator: DocumentPickerCoordinatorDelegate {
                         break
                     }
                 }
-
+                
                 if coordinator.currentPickerDismissesAutomatically {
                     self.cameraScreen?.showErrorDialog(for: error,
                                                        positiveAction: positiveAction)
                 } else {
                     coordinator.currentPickerViewController?.showErrorDialog(for: error,
-                                                                    positiveAction: positiveAction)
+                                                                             positiveAction: positiveAction)
                 }
             }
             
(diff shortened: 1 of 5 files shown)


Example/Example Swift/ComponentAPICoordinator.swift
```diff
@@ -409,15 +409,25 @@ extension ComponentAPICoordinator: CameraViewControllerDelegate {
...
             switch result {
             case .success(let documentRequests):
-                self.documentRequests.append(contentsOf: documentRequests)
-                self.upload(documentRequests: documentRequests)
-                
-                if self.giniConfiguration.multipageEnabled, self.documentRequests.count > 1 {
-                    if let imageDocument = document as? GiniImageDocument {
+                switch document {
+                case let qrCodeDocument as GiniQRCodeDocument:
+                    viewController.showPopup(forQRDetected: qrCodeDocument) {
```
nsobadjiev: Wouldn't it be better to show the popup before the document is validated and definitely before the `cameraDidCapture` method is invoked? That would allow you to keep this code generic and handle all document types the same way.

Example/Example Swift/ComponentAPICoordinator.swift
```diff
@@ -409,15 +409,25 @@ extension ComponentAPICoordinator: CameraViewControllerDelegate {
         validate([document]) { result in
             switch result {
             case .success(let documentRequests):
-                self.documentRequests.append(contentsOf: documentRequests)
-                self.upload(documentRequests: documentRequests)
-                
-                if self.giniConfiguration.multipageEnabled, self.documentRequests.count > 1 {
-                    if let imageDocument = document as? GiniImageDocument {
+                switch document {
```
nsobadjiev: I'm not sure a switch is appropriate here. It's just a special case for a QR code. Everything else is the same (I assume a PDF document cannot appear here). If we add another type, it would fall through to the break statement. That's why I think a simple if statement would be better.

Example/Example Swift/ComponentAPICoordinator.swift
```diff
@@ -520,6 +530,7 @@ extension ComponentAPICoordinator: MultipageReviewViewControllerDelegate {
             
             if self.giniConfiguration.multipageEnabled, self.documentRequests.type == .image {
                 self.refreshMultipageReview(with: self.documentRequests)
+
```
nsobadjiev: We don't need this extra line?

GiniVision/Classes/Core/GiniScreenAPICoordinator+Camera.swift
```diff
@@ -25,20 +25,28 @@ extension GiniScreenAPICoordinator: CameraViewControllerDelegate {
...
+                    }
+                case let imageDocument as GiniImageDocument:
+                    self.addToDocuments(new: [validatedDocument])
+                    self.didCaptureAndValidate(document)
+                        
+                    // In case that there is more than one image already captured, an animation is shown instead of
+                    // going to next screen
                     if self.documentRequests.count > 1 {
                         viewController.animateToControlsView(imageDocument: imageDocument)
+                        return
```
nsobadjiev: Do you really need this return?

Example/Example Swift/ComponentAPICoordinator.swift
kikettas: No, because you don't show the popup for not valid QRCodes. If they don't have the right format (bezahle, stuzza...), it shouldn't be shown.

Example/Example Swift/ComponentAPICoordinator.swift
kikettas: You are right, and `ifelse` would be better here.

GiniVision/Classes/Core/GiniScreenAPICoordinator+Camera.swift
kikettas: No, good catch. 👍

Example/Example Swift/ComponentAPICoordinator.swift
```diff
@@ -409,15 +409,22 @@ extension ComponentAPICoordinator: CameraViewControllerDelegate {
         validate([document]) { result in
             switch result {
             case .success(let documentRequests):
-                self.documentRequests.append(contentsOf: documentRequests)
-                self.upload(documentRequests: documentRequests)
-                
-                if self.giniConfiguration.multipageEnabled, self.documentRequests.count > 1 {
-                    if let imageDocument = document as? GiniImageDocument {
+                if let qrCodeDocument = document as? GiniQRCodeDocument {
+                    viewController.showPopup(forQRDetected: qrCodeDocument) {
```
nsobadjiev: Did you consider showing the popup before invoking the `didCapture` method.
I really think it should be part of the camera view controller's responsibilities to return a QR code only after it has been confirmed. It would also make the flow of the code clearer.

Example/Example Swift/MultipageDocumentService.swift
```diff
@@ -15,16 +15,28 @@ final class MultipageDocumentsService: DocumentServiceProtocol {
...
         let partialDocumentsInfoSorted = partialDocuments
+            .lazy
             .map { $0.value }
             .sorted()
             .map { $0.info }
+        
+        // When a PDF/QrCode document is imported the analysis screen is shown right away, and therefore the analysis
+        // is triggered. There could be the case where the document hadn't been analyzed when this happens,
+        // that's why a reference to the completion block has to be kept. Once the document is uploaded,
+        // the completion block is called (see below).
```
nsobadjiev: This comment is really confusing. It's 4 lines and still doesn't make things more clear. Can you explain the different steps of uploading and polling for extractions why how the issue might arise?

Example/Example Swift/MultipageDocumentService.swift
```diff
@@ -15,16 +15,28 @@ final class MultipageDocumentsService: DocumentServiceProtocol {
     var partialDocuments: [String: PartialDocumentInfo] = [:]
     var compositeDocument: GINIDocument?
     var analysisCancellationToken: BFCancellationTokenSource?
-    
+    var pendingAnalysisHandler: AnalysisCompletion?
```
nsobadjiev: I am really against this. It just makes me think it's just a patch to make things somehow work. The Document service should know how to manage a document's lifecycle. Keep track if it's uploaded, if it makes sense to poll for extractions, etc. The only input it needs (from the UI) is if anyone is still interested in the result. And even in this case, I'd say it's better to follow the straight case and provide a way for the UI to say that it's no longer interested in the result and cancel.

In my experience, holding on to such closures because something might get delayed brings more problems than solutions.

GiniVision/Classes/Core/CameraViewController.swift
```diff
@@ -487,7 +487,7 @@ extension CameraViewController {
      - parameter qrDocument: A validated `GiniQRCodeDocument`.
      - parameter didTapDone: Block executed when the user taps the _Done_ button to proceed with the `GiniQRCodeDocument`.
      */
-    func showPopup(forQRDetected qrDocument: GiniQRCodeDocument, didTapDone: @escaping () -> Void) {
+    public func showPopup(forQRDetected qrDocument: GiniQRCodeDocument, didTapDone: @escaping () -> Void) {
```
nsobadjiev: If you make the popup part of CameraViewController's internal responsibilities, you'd be able to remove this `public`. Which would be awesome since it's a confusing part of the public interface. What popup? Why?

Example/Example Swift/ComponentAPICoordinator.swift
kikettas: I decided to use this approach because the validation responsibility should be unified outside of the camera screen. 
What I may consider is to validate the QR codes before delivering them to the camera, seeing if it is worth in terms of performance.
Do you what me to tackle this in a different PR?

GiniVision/Classes/Core/CameraViewController.swift
kikettas: Answered above.

Example/Example Swift/MultipageDocumentService.swift
kikettas: #241

Example/Example Swift/ComponentAPICoordinator.swift
kikettas: #242

PR #244, opened 2018-06-07
Added analysis screen error view

kikettas:
When there is an error in the analysis or uploading a document, it has to be shown in the analysis screen. For the uploading doesn't apply for images when multipage is enabled, since the error will be shown in the multipage screen.
Also fixed minor issue with auto-layout in page cell.

### How to test
Run the Example app, disable the internet connection, scan a qr code and see that an error is shown in the analysis screen.

### Merging
Automatic

### Issue
#243

### DIFF
--- a/Example/Example Swift/ComponentAPICoordinator.swift
+++ b/Example/Example Swift/ComponentAPICoordinator.swift
@@ -302,7 +302,9 @@ extension ComponentAPICoordinator {
                     if self.documentRequests.type != .image || !self.giniConfiguration.multipageEnabled {
                         guard let visionError = error as? GiniVisionError,
                         let firstDocumentRequest = self.documentRequests.first else { return }
-                        self.showErrorInAnalysisScreen(with: visionError.message, for: firstDocumentRequest)
+                        self.showErrorInAnalysisScreen(with: visionError.message, for: firstDocumentRequest) {
+                            self.upload(documentRequest: firstDocumentRequest)
+                        }
                     }
                 }
                 
@@ -325,7 +327,9 @@ extension ComponentAPICoordinator {
                     guard let firstDocumentRequest = self.documentRequests.first else { return }
                     let visionError = CustomAnalysisError.analysisFailed
 
-                    self.showErrorInAnalysisScreen(with: visionError.message, for: firstDocumentRequest)
+                    self.showErrorInAnalysisScreen(with: visionError.message, for: firstDocumentRequest) {
+                        self.startAnalysis()
+                    }
                 }
             }
         })
@@ -335,14 +339,14 @@ extension ComponentAPICoordinator {
         documentService?.delete(document)
     }
     
-    private func showErrorInAnalysisScreen(with message: String, for documentRequest: DocumentRequest) {
+    private func showErrorInAnalysisScreen(with message: String,
+                                           for documentRequest: DocumentRequest,
+                                           action: @escaping () -> Void) {
         if self.analysisScreen == nil {
             self.analysisScreen = AnalysisViewController(document: documentRequest.document)
         }
         
-        self.analysisScreen?.showError(with: message, action: {
-            self.upload(documentRequest: documentRequest)
-        })
+        self.analysisScreen?.showError(with: message, action: action)
     }
 }
 
--- a/Example/Example Swift/ComponentAPICoordinator.swift
+++ b/Example/Example Swift/ComponentAPICoordinator.swift
@@ -249,7 +249,12 @@ extension ComponentAPICoordinator {
     
     fileprivate func push<T: UIViewController>(viewController: UIViewController, removing viewControllers: [T?]) {
         var navigationStack = navigationController.viewControllers
-        let viewControllersToDelete = navigationStack.filter { return viewControllers.map { $0.self }.contains($0) }
+        let viewControllersToDelete = navigationStack.filter {
+            return viewControllers
+                .lazy
+                .compactMap { $0 }
+                .contains($0)
+        }
         
         viewControllersToDelete.forEach { viewControllerToDelete in
             if let index = navigationStack.index(of: viewControllerToDelete) {
--- a/Example/Example Swift/ComponentAPICoordinator.swift
+++ b/Example/Example Swift/ComponentAPICoordinator.swift
@@ -197,7 +197,7 @@ extension ComponentAPICoordinator {
                                                       action: #selector(closeComponentAPIFromResults))
         }
         
-        push(viewController: resultsScreen!, removingViewControllerOfType: AnalysisViewController.self)
+        push(viewController: resultsScreen!, removing: [reviewScreen, analysisScreen])
     }
     
     fileprivate func showNoResultsScreen() {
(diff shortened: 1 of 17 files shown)


Example/Example Swift/ComponentAPICoordinator.swift
```diff
@@ -171,8 +171,11 @@ extension ComponentAPICoordinator {
     @objc fileprivate func showAnalysisScreen() {
         guard let document = documentRequests.first?.document else { return }
         
+        if analysisScreen == nil {
```
nsobadjiev: Is there a valid case when the analysis screen might not be nil here?
Later on in the method you push it to the navigation stack. If the analysis screen is already existing, it is also quite likely that it's somewhere in the view hierarchy. If you try to push it (again), it will fail/crash/burn.

Example/Example Swift/ComponentAPICoordinator.swift
```diff
@@ -244,13 +245,21 @@ extension ComponentAPICoordinator {
         closeComponentAPI()
     }
     
-    fileprivate func push<T>(viewController: UIViewController, removingViewControllerOfType: T.Type) {
+    fileprivate func push<T: UIViewController>(viewController: UIViewController, removing viewControllers: [T?]) {
```
nsobadjiev: It might be actually better to make it an extension to the `UINavigationController`. But just a suggestion.

Example/Example Swift/ComponentAPICoordinator.swift
```diff
@@ -287,6 +296,13 @@ extension ComponentAPICoordinator {
                     self.documentRequests[index].isUploaded = true
                 case .failure(let error):
                     self.documentRequests[index].error = error
+                    
+                    if self.documentRequests.type != .image || !self.giniConfiguration.multipageEnabled {
+                        guard let visionError = error as? GiniVisionError else { return }
+                        self.showErrorInAnalysisScreen(with: visionError.message, for: documentRequest) {
+                            self.upload(documentRequest: documentRequest)
```
nsobadjiev: So, the `action` closure here is escaping. The method is called on the weak `self` from the outer closure and the inner closure also uses the same `self`. Are you sure that doesn't create a retain cycle?

Example/Example Swift/ComponentAPICoordinator.swift
```diff
@@ -317,6 +335,16 @@ extension ComponentAPICoordinator {
     fileprivate func delete(document: GiniVisionDocument) {
         documentService?.delete(document)
     }
+    
+    private func showErrorInAnalysisScreen(with message: String,
+                                           for documentRequest: DocumentRequest,
+                                           action: @escaping () -> Void) {
+        if self.analysisScreen == nil {
+            self.analysisScreen = AnalysisViewController(document: documentRequest.document)
```
nsobadjiev: If you have to create an analysis screen, don't you also have to put in on screen?

Example/Example Swift/MultipageDocumentService.swift
```diff
@@ -27,6 +27,7 @@ final class MultipageDocumentsService: DocumentServiceProtocol {
             .map { $0.value }
             .sorted()
             .map { $0.info }
+            .filter { $0.documentUrl != nil}
```
nsobadjiev: If they don't have a `documentUrl`, they are not uploaded yet?

GiniVision/Classes/Core/AnalysisViewController.swift
```diff
@@ -194,6 +203,22 @@ import UIKit
         loadingIndicatorView.stopAnimating()
     }
     
+    /**
+     Shows an error when there was an error with either during the analysis or uploading a document
```
nsobadjiev: Shows an error when there was an error with either <strike>during</strike> the analysis or **document upload**

Example/Example Swift/ComponentAPICoordinator.swift
kikettas: If it were for the Screen API, I would do it. But I don't see an advantage of creating it for the Component API.

Example/Example Swift/ComponentAPICoordinator.swift
kikettas: I'm not super convince about this thing with the `AnalysisViewController`, but the other thing would be to keep a reference to the analysis error and passing it to the view controller which I don't like either although maybe it is better. What do you think?

Example/Example Swift/ComponentAPICoordinator.swift
kikettas: When showing an error, I just answered you about it below.

Example/Example Swift/ComponentAPICoordinator.swift
kikettas: No, it doesn't 👍

Example/Example Swift/MultipageDocumentService.swift
kikettas: Exactly, that it might change if I find a solution or the pending handler.

Example/Example Swift/ComponentAPICoordinator.swift
nsobadjiev: What's the difference?

GiniVision/Classes/Core/GiniScreenAPICoordinator+Analysis.swift
```diff
@@ -67,13 +67,18 @@ extension GiniScreenAPICoordinator: AnalysisDelegate {
...
-                let action = action,
-                let firstDocument = self.documentRequests.first?.document else { return }
-            if self.analysisViewController == nil {
-                self.analysisViewController = self.createAnalysisScreen(withDocument: firstDocument)
-            }
+                let action = action else { return }
             
-            self.analysisViewController?.showError(with: message, action: action)
+            if let analysisViewController = self.analysisViewController {
+                analysisViewController.showError(with: message, action: { [weak self] in
```
nsobadjiev: Is code is a duplicate to the one in the component API coordinator. In general a lot of code is duplicated between the two coordinators. I know that they are different ways to use the library and they have to do the same thing most of the time, but it would still be nice to combine what you can. I constantly see you fix something in the screen API and then remember to transfer it also to the component API. It's error prone.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -42,6 +42,11 @@ final class GiniScreenAPICoordinator: NSObject, Coordinator {
     fileprivate(set) var documentRequests: [DocumentRequest] = []
     fileprivate let multiPageTransition = MultipageReviewTransitionAnimator()
     weak var visionDelegate: GiniVisionDelegate?
+    
+    // When there was an error uploading a document or analyzing it and the analysis screen is not initialized yet,
```
nsobadjiev: If have too keep the tense between the two sub-sentenses. If one has "was" the other needs to also be in past tense. For instance:

> When there was an error uploading a document or analyzing it and the analysis screen hadn't been initialized yet

Also applies to the other comment in the component API, which is a copy of this one.

GiniVision/Classes/Core/GiniScreenAPICoordinator+Analysis.swift
kikettas: I know that is prone error, but they have to be different since they are different approaches. In this case I think it cannot be shared, but I always try to share the most code I can (since it is easier when implementing)

Issue #236, opened 2018-06-04
Fix issue with lazy loading for bundle data in tests


kikettas:
When using lazy loading in tests, they are loaded several times instead of only one, leading to perfomance issues when loading data from the test bundle.

PR #246, opened 2018-06-11
Improve tests performance

kikettas:
Unit tests performance were lasting almost 20 second when they shouldn't. The reasons of this bad performance were:

- Image documents weren't lazy loaded in the `MultipageReviewViewControllerTest` class, being reloaded for every test case.
- Component API tests were using the `DocumentService` implementation instead of a mocked one.
- Select cell expectation in `GalleryCoordinatorTests` cases was asynchronous when actually the selection wasn't

### How to test
Run the tests suite an see that they last ~5 seconds instead of ~20 seconds.

### Merging
Automatic

### Issue
#236

### DIFF
--- a/Example/Example Swift/ComponentAPIDocumentServiceProtocol.swift
+++ b/Example/Example Swift/ComponentAPIDocumentServiceProtocol.swift
@@ -1,5 +1,5 @@
 //
-//  DocumentServiceProtocol.swift
+//  ComponentAPIDocumentServiceProtocol.swift
 //  GiniVision
 //
 //  Created by Enrique del Pozo Gómez on 3/29/18.
@@ -21,30 +21,30 @@ enum CustomAnalysisError: GiniVisionError {
     }
 }
 
-enum Result<T> {
+enum CompletionResult<T> {
     case success(T)
     case failure(Error)
 }
 
-typealias UploadDocumentCompletion = (Result<GINIDocument>) -> Void
-typealias AnalysisCompletion = (Result<[String: Extraction]>) -> Void
+typealias ComponentAPIUploadDocumentCompletion = (CompletionResult<GINIDocument>) -> Void
+typealias ComponentAPIAnalysisCompletion = (CompletionResult<[String: Extraction]>) -> Void
 
-protocol DocumentServiceProtocol: class {
+protocol ComponentAPIDocumentServiceProtocol: class {
     
     var giniSDK: GiniSDK { get }
     var compositeDocument: GINIDocument? { get set }
     var analysisCancellationToken: BFCancellationTokenSource? { get set }
     
     init(sdk: GiniSDK)
-    func startAnalysis(completion: @escaping AnalysisCompletion)
+    func startAnalysis(completion: @escaping ComponentAPIAnalysisCompletion)
     func cancelAnalysis()
     func delete(_ document: GiniVisionDocument)
     func upload(_ document: GiniVisionDocument,
-                completion: UploadDocumentCompletion?)
+                completion: ComponentAPIUploadDocumentCompletion?)
     func update(_ imageDocument: GiniImageDocument)
 }
 
-extension DocumentServiceProtocol {
+extension ComponentAPIDocumentServiceProtocol {
     
     var rotationDeltaKey: String { return "rotationDelta" }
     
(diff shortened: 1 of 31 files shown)


Example/Example Swift/ComponentAPIDocumentServiceProtocol.swift
```diff
@@ -21,30 +21,30 @@ enum CustomAnalysisError: GiniVisionError {
...
     case failure(Error)
 }
 
-typealias UploadDocumentCompletion = (Result<GINIDocument>) -> Void
-typealias AnalysisCompletion = (Result<[String: Extraction]>) -> Void
+typealias ComponentAPIUploadDocumentCompletion = (CompletionResult<GINIDocument>) -> Void
+typealias ComponentAPIAnalysisCompletion = (CompletionResult<[String: Extraction]>) -> Void
 
-protocol DocumentServiceProtocol: class {
+protocol ComponentAPIDocumentServiceProtocol: class {
```
nsobadjiev: Why shouldn't the document service protocol be the same for the component and screen API? What's the difference?

Example/Tests/ComponentAPICoordinatorTests.swift
```diff
@@ -9,17 +9,23 @@
...
-class GINIComponentAPICoordinatorTests: XCTestCase {
+final class ComponentAPICoordinatorTests: XCTestCase {
     
     var componentAPICoordinator: ComponentAPICoordinator?
+
+    var documentService: ComponentAPIDocumentServiceProtocol = {
+        let sdkBuilder = GINISDKBuilder.anonymousUser(withClientID: "",
+                                                      clientSecret: "",
+                                                      userEmailDomain: "")
+       return DocumentServiceMock(sdk: sdkBuilder!.build()!)
```
nsobadjiev: If it's a mock, why does it even require an SDK instance? It seems redundant and requires you to write some extra code above.

Example/Tests/GiniVisionTestsHelper.swift
```diff
@@ -1,50 +1,47 @@
...
+//  Created by Enrique del Pozo Gómez on 6/5/18.
+//  Copyright © 2018 Gini GmbH. All rights reserved.
 //
 
 import Foundation
-import UIKit
-import XCTest
 @testable import GiniVision
 
-internal extension XCTestCase {
```
nsobadjiev: Why can't this still be an extension to the `XCTestCase` class? It was more convenient to call methods on self and not on a separate class. 
And how does having class methods improve performance?

Example/Example Swift/ComponentAPIDocumentServiceProtocol.swift
kikettas: I had to rename it because the compiler wasn't able to see that they were different and I was having problems with the test target.

Example/Tests/GiniVisionTestsHelper.swift
kikettas: When fetching the images/pdfs from the bundle an instance of `self` had to be provided, being a problem when loading these files in a non-lazy way.
I decided to do it in a separate class because at the end, although it was more convenient and in terms of performance is the same, it is not the responsibility of the TestCase to load a file.

What improved was not the class methods but not loading the same file several times (there was a problem with `lazy`properties).

Example/Tests/ComponentAPICoordinatorTests.swift
kikettas: Because both the initializer and the stored property are part of the protocol

nsobadjiev: You don't have to follow the protocol since you are referring to the concrete subclass. You can use a convenience initializer to create and mock and then just use that it conforms to the document service protocol

Example/Tests/GiniVisionTestsHelper.swift
nsobadjiev: I understand the performance improvements, but now the code is not a easy to read and write.

Example/Tests/ComponentAPICoordinatorTests.swift
kikettas: You are right, that could be an option.

Example/Tests/DocumentServiceMock.swift
```diff
@@ -42,3 +42,9 @@ final class DocumentServiceMock: ComponentAPIDocumentServiceProtocol {
     }
 
 }
+
+extension DocumentServiceMock {
+    convenience init() {
```
nsobadjiev: I don't see a need for this to be in an extension. But as you prefer.

PR #248, opened 2018-06-13
Added customizable opaque background view style when the tool tip is shown 

kikettas:
The background view color when the tool tip was shown in the camera screen might be misleading for the user. Now it is possible to customize it with a different style: 

- Blurred with light style
- Blurred with dark style
- Dimmed (only dark)

### How to test
Change the value of `GiniConfiguration.toolTipOpaqueBackgroundStyle` and run the app for the first time. Open the screen API and check that the background shown below the tool tip matches the value specified in the property.

### Merging
Automatic

### Issue
#247

### DIFF
--- a/GiniVision/Classes/OpaqueViewFactory.swift
+++ b/GiniVision/Classes/OpaqueViewFactory.swift
@@ -12,9 +12,9 @@ public enum OpaqueViewStyle {
     case dimmed
 }
 
-final class OpaqueViewFactory {
+struct OpaqueViewFactory {
     
-    class func create(with style: OpaqueViewStyle) -> UIView {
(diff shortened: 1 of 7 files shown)


GiniVision/Classes/OpaqueViewFactory.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+
+import UIKit
+
+public enum OpaqueViewStyle {
+    case blurred(style: UIBlurEffectStyle)
+    case dimmed
+}
+
+final class OpaqueViewFactory {
```
a-szotyori: Could be a struct.

Issue #241, opened 2018-06-07
Remove handler for incomplete document uploads


kikettas:
The `pendingAnalysisHandler` could lead to big issues in the future, so it will be removed being handled in the UI

PR #254, opened 2018-06-19
Remove pending handler

kikettas:
### Description
In order to simplify and clarify the networking in the GVL, the pending handler in the document service has been removed. With this, both the Multi page and Single page document services have been unified into the same class (for both the Component and Screen APIs). This forces the document upload once it has been reviewed in the single page mode instead of doing it once the image has been captured. This would only affect the single page mode.

### How to test
Run the example app, and try to analyze a document in both single and multi page modes

### Merging
Automatic

### Issues
#241


Example/Example Swift/ComponentAPICoordinator.swift
```diff
@@ -392,23 +413,30 @@ extension ComponentAPICoordinator: UINavigationControllerDelegate {
...
+            if operation == .pop {
+                documentService?.cancelAnalysis()
+            }
         }
         
-        if toVC is CameraViewController && (fromVC is ReviewViewController || fromVC is AnalysisViewController) {
+        if toVC is CameraViewController &&
+            (fromVC is ReviewViewController ||
+             fromVC is AnalysisViewController ||
+             fromVC is ImageAnalysisNoResultsViewController) {
```
nsobadjiev: Hmm, so all this condition does is check if the app is in "Single page" mode (as opposed to multi page) and the user is cancelling the processing?

If so, it is a little fragile. Instead of knowing that it's "Single page", you check for all possible view controller configurations where that's the case. There's a danger that if you change something in the screen flow, this will no longer be the right check.

Example/Example Swift/ComponentAPIDocumentsService.swift
```diff
@@ -0,0 +1,259 @@
...
+                if let error = task?.error {
+                    let id = self.compositeDocument?.documentId ?? ""
+                    let message = "❌ Error sending feedback for document with id: \(id) error: \(error)"
+                    print(message)
+                    
+                    return nil
+                }
+                
+                print("🚀 Feedback sent with \(updatedExtractions.count) extractions")
+
```
nsobadjiev: Is there a reason for all these spaces?

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
```diff
@@ -234,24 +237,28 @@ extension GiniScreenAPICoordinator: UINavigationControllerDelegate {
...
             if let firstDocument = pages.first?.document {
                 visionDelegate?.didCancelReview(for: firstDocument)
             }
         }
         
-        if toVC is CameraViewController && (fromVC is ReviewViewController || fromVC is AnalysisViewController) {
+        if toVC is CameraViewController &&
+            (fromVC is ReviewViewController ||
+             fromVC is AnalysisViewController ||
+             fromVC is ImageAnalysisNoResultsViewController) {
```
nsobadjiev: Same remark about checking for "Single Page".

Example/Example Swift/ComponentAPICoordinator.swift
kikettas: It is not only Single page but also when a PDF or a QRCode is analyzed in Multi page mode.
For me the `navigationController(_ navigationController:operation:fromVC:toVC:)` method is the key in the navigation since every transition can be intercepted, so this must be done here and not in the `backTo` or `goTo` actions of every screen. It is still a little bit fragile but less than the other cases, and makes sense to do it in this method.

GiniVision/Classes/Core/GiniScreenAPICoordinator.swift
kikettas: Answered above.

Example/Example Swift/ComponentAPICoordinator.swift
nsobadjiev: I'm not questioning whether or not the code should be here, but the way the condition is checked. It's fragile to look into the combination of view controllers and how they transition from one to another.

It would have been better for the Coordinator to already know if it's the single or multipage mode (or the type of document) and knew what to do without deriving the current situation purely based to the screen transitions.

kikettas: That's the first think I thought, but still need to know what is the current screen, which could lead to some spaghetti code (when every screen is shown/removed)

PR #255, opened 2018-06-22
Last tweaks and fixes

kikettas:
### Description
Following issues have been fixed:

- Delete button wasn't disabled when a deletion was being performed.
- Cancelled analysis was shown as an error
- Mixed collections weren't taken into account when using QR codes.
- Old `CameraViewController` initializers should be marked as `unavailable` and not as `deprecated` since there is a breaking change after taking the DocumentPickerCoordinator out of this screen.
- The cancellation token wasn't passed to the `getExtractions` method, which made the analysis not cancellable.
- `DocumentService` was used in the `GiniScreenAPICoordinator` class when it shouldn't. (networking must be always handled outside of the `Core` module)

And also this minor additions where integrated:

- Missing views styling in customization guide.
- Improved performance in `MultipageReviewViewController` when it appears (collections are now reloaded in parallel and without animations)

### Merging
Automatic


Documentation/source/Customization guide.md
```diff
@@ -132,8 +136,12 @@ Customizable assets can be found in [the Assets repo](https://github.com/gini/gi
 	 - Image &#8594; <span style="color:#009EDF">*navigationAnalysisBack*</span> image asset
 	 - Title &#8594; <span style="color:#009EDF">*ginivision.navigationbar.analysis.back*</span> localized string
 
-##### 2. Loading view
-- Indicator color &#8594; `GiniConfiguration.analysisLoadingIndicatorColor`
+##### 3. PDF Information view
+- Text color &#8594; `GiniConfiguration.analysisPDFInformationTextColor`
+- Text color &#8594; `GiniConfiguration.analysisPDFInformationBackgroundColor`
```
nsobadjiev: Text color? The code snippet says "BackgroundColor".

Documentation/source/Customization guide.md
```diff
@@ -132,8 +136,12 @@ Customizable assets can be found in [the Assets repo](https://github.com/gini/gi
...
+
+##### 3. Loading view
```
nsobadjiev: After 3 comes...?
1. 4
2. another 3

PR #265, opened 2018-07-05
Fixes and changes from testing

kikettas:
### Description
After the testing session, some fixes and minor changes were introduced. These are:

- Rearrange toolbar items in multipage screen
- Fixed issue with UIDocumentPickerViewController in iOS<11.0 (a method must be implemented, Xcode didn't warn about it)
- Changed collection type in `GalleryCoordinator`, so the items are retrieved in order.
- Added custom opaque view in multipage screen.
- Fixed issue with log in iOS<10.0.
- Fixed breaking constraints in small devices.
- The session is retrieved when the SDK is created.

### Merging
Automatic


kikettas:
[Fix issues from user testing](https://trello.com/c/7IlKAkxU/1-fix-issues-from-user-testing)

kikettas:
[Fix issues from user testing](https://trello.com/c/7IlKAkxU/1-fix-issues-from-user-testing)

kikettas:
[Fix issues from user testing](https://trello.com/c/7IlKAkxU/2-fix-issues-from-user-testing)

GiniVision/Classes/Core/Multipage Review/MultipageReviewViewController.swift
```diff
@@ -392,9 +395,9 @@ extension MultipageReviewViewController {
     }
     
     fileprivate func createReorderPagesTip() {
-        blurEffect = UIVisualEffectView(effect: UIBlurEffect(style: .light))
-        blurEffect?.alpha = 0
-        self.view.addSubview(blurEffect!)
+        opaqueView = OpaqueViewFactory.create(with: giniConfiguration.multipageToolTipOpaqueBackgroundStyle)
+        opaqueView?.alpha = 0
+        self.view.addSubview(opaqueView!)
```
nsobadjiev: You can avoid this force unwrapping if you work with a local variable, set it up and at the end of the method assign `opaqueView` to it. But not super important.

GiniVision/Classes/Networking/DocumentsService.swift
```diff
@@ -17,6 +17,7 @@ final class DocumentsService: DocumentServiceProtocol {
     
     init(sdk: GiniSDK) {
         self.giniSDK = sdk
+        _ = self.getSession()
```
nsobadjiev: Not really a fan of that. Methods shouldn't just be call for their side effects.
We do it in tests when we call `_ = controller.view` but that's different - it's not production code.

GiniVision/Classes/Networking/DocumentsService.swift
```diff
@@ -17,6 +17,7 @@ final class DocumentsService: DocumentServiceProtocol {
     
     init(sdk: GiniSDK) {
         self.giniSDK = sdk
+        self.giniSDK.sessionManager.logIn()
```
nsobadjiev: I don't want to be this guy, but do you think the constructor is a suitable place for something like a login?
In my opinion, a construction should only init the object and not start such tasks/side effects.
It might be inconvenient for users.

PR #266, opened 2018-07-05
Consider all localization keys in app bundle

siegeral:
### Description
Allow all defined localizable string keys to be overwritten in client's bundle

### How to test
Overwrite localizable string keys in Localizable.strings

### Merging
Automatic


kikettas:
Hi @siegeral!

First, thank you for contributing to this repository 👍! I reviewed the changes you want to integrate in the Gini Vision Library (making all strings customizable outside of the `GiniVision` bundle) and I have to tell you that unfortunately they cannot be considered. Before releasing the GVL 3.2.0, we carried out several UX and UI testing sessions where we tried out different approaches in order to have the best user experience and that's why we decided to keep some functionality inmutable.

Nonetheless, I recommend you to write Gini [technical support](mailto:technical-support@gini.net) if you have any further question.

siegeral:
Ok, acknowledged.

PR #268, opened 2018-07-06
Fix german texts

kikettas:
### Description
Fixed issues with german texts.

### How to test
Run the app and check that the new texts are added.

### Merging
After #265


kikettas:
[Fix issues from user testing](https://trello.com/c/7IlKAkxU/2-fix-issues-from-user-testing)

PR #281, opened 2018-08-01
Extract string localizations into enum

kikettas:
### Description
To make it safer and less prone to errors, all the localizable strings have been extracted into different enums.

### How to test
Run the example app and check that all the texts are localized.

### Merging
Automatic


kikettas:
[Extract strings and images resources into an enum](https://trello.com/c/tzszhpYg/83-extract-strings-and-images-resources-into-an-enum)

Example/Tests/SupportedFormatsViewControllerTests.swift
```diff
@@ -15,29 +15,29 @@ final class SupportedFormatsViewControllerTests: XCTestCase {
     let initialGiniConfiguration = GiniConfiguration.shared
     
     var sections: [SupportedFormatCollectionSection] = [
-        (NSLocalizedString("ginivision.supportedFormats.section.1.title",
```
nsobadjiev: I don't understand why you don't use the same `.localized` method thing.
If it's because you also want to check is the proper, localized string is used, why not just used the actual string people would see at runtime?

GiniVision/Classes/Core/Custom views/NoticeView.swift
```diff
@@ -14,11 +14,9 @@ import UIKit
     var title: String {
         switch self {
         case .retry:
-            return NSLocalizedStringPreferred("ginivision.multipagereview.error.retryAction",
-                                               comment: "button title for retry action")
+            return .localized(resource: MultipageReviewStrings.retryActionButton)
```
nsobadjiev: You moved the comment to the *Strings enum. I feel that you lost some additional hint here and in the enum, it's not really helpful. It would be better to see where hints at the place these strings are used (hence, here).

I wouldn't say that these comments are really critical, but since we already have them, I'd rather keep them at the place where they are used.

GiniVision/Classes/Core/GiniConfiguration.swift
```diff
@@ -616,11 +574,7 @@ import UIKit
      - Warning: Deprecated, use the ginivision.review.bottom localized string instead
 
      */
-    @objc public var reviewTextBottom =
-        NSLocalizedStringPreferred("ginivision.review.bottom",
-                                   comment: "Text at the bottom of the review screen encouraging the " +
-                                            "user to check sharpness by double-tapping the image")
-    
+    @objc public var reviewTextBottom: String = .localized(resource: ReviewStrings.bottomText)
```
nsobadjiev: I suppose that there are two equivalent ways to call these methods:

`reviewTextBottom: String = .localized` and
`reviewTextBottom = String.localized`

In my opinion the second is a bit better. But maybe just my preference.

GiniVision/Classes/Core/Helpers/Localization/LocalizableStringResource.swift
```diff
@@ -0,0 +1,34 @@
...
+    var tableName: String { get }
+    var tableEntry: LocalizationEntry { get }
+    var customizable: Bool { get }
+    var args: CVarArg? { get }
+
+}
+
+extension LocalizableStringResource {
+    var localized: String {
+        let key = "ginivision.\(tableName).\(tableEntry.value)"
```
nsobadjiev: I could argue that even though you are now using enums, the whole localization is still not type safe. In this line, you are still relying on constructing strings so they match a text file. If you typed the table name or entry wrong, the compiler wouldn't complain, but it would rather fail at runtime.

GiniVision/Classes/Core/Helpers/Localization/LocalizableStringResource.swift
```diff
@@ -0,0 +1,34 @@
...
+
+import Foundation
+import UIKit
+
+typealias LocalizationEntry = (value: String, description: String)
+
+protocol LocalizableStringResource {
+    var tableName: String { get }
+    var tableEntry: LocalizationEntry { get }
+    var customizable: Bool { get }
```
nsobadjiev: I could argue that this is not type safe. When you add a new entry to the specific enums, you will have to remember to  also set the `customizable` and possibly the `args` fields. If you forget to do so, the default value would be used, which would often be undesirable.

Since adding new strings is expected, I'd consider this an issue. I'd try to have these fields as associated types so every time someone adds a string, the compiler "reminds" them to also provide a value for the `customizable` and `args` field.

GiniVision/Classes/Core/Custom views/NoticeView.swift
kikettas: I thought about it and I found better to have them separately, mostly for those times when there is more than one localized strings in a row.  In makes the code more readable and clear (most of the times you have to split the comment into 2-3 lines because it does not fit the line width)

GiniVision/Classes/Core/GiniConfiguration.swift
kikettas: It could be better for those situations when you are forced to add it to the property to resolve the ambiguity, but there are a lot of different situations where it is not needed. For example, when assigning an already defined `String` property (i.e: `title` in a `UIViewController`)

GiniVision/Classes/Core/Helpers/Localization/LocalizableStringResource.swift
kikettas: But do you know a better way for making it safer? Because I'm not sure that I can find a better and non-complex approach for that. I know that there safer approaches like `R.swift` library, but in my opinion that's an overkill.

GiniVision/Classes/Core/Helpers/Localization/LocalizableStringResource.swift
kikettas: I agree with you that this could lead to an issue in the future, but I don't get what you mean in the last paragraph. Every `LocalizableStringResource` implementation is a bunch of strings, and I think it is not possible to add an associatetype for each case.

Nonetheless, I think that removing the `default` declaration inside of `customizable` switch would be enough to make it more visible when adding a new string.

nsobadjiev: The case I don't like is the following:
* I want to add another string
* I add it to one of the enums
* I don't pay attention to the `customizable` field so I don't do anything with it

In this case, my new string uses the default value which might be an error that will only show up at runtime.

I would have been nice to have type safety on that and now allow new enum values without explicitly specifying a `customizable` value.

kikettas: But how can you add type safety to an enum case? What I have is to remove the `default` implementation, so every time you add a new string in the resource, you will have to specify not only the tableEntry but also if it is customizable or not, otherwise it won't compile.

Example/Tests/SupportedFormatsViewControllerTests.swift
kikettas: You are right, makes sense to use the same method for the tests.

PR #283, opened 2018-09-11
Feature move tests to subspec

kikettas:
### Description
Moved all GiniVision tests into a test subspec. The tests related to the Example app have been kept in the same target (renamed from `GiniVision_Tests` to `Example_Tests`).
Some method's signature has been also improved.

### How to test
Run the job in Jenkins and check that there are two test target running (one for the GiniVision librbary and the other for the Example app).

### Merging
Automatic


GiniVision/Tests/GiniVisionTestsHelper.swift
```diff
@@ -0,0 +1,61 @@
...
+import UIKit
+@testable import GiniVision
+
+final class GiniVisionTestsHelper {
+    
+    class func fileData(named name: String, fileExtension: String) -> Data? {
+        return try? Data(contentsOf: urlFromFile(named: name, fileExtension: fileExtension)!)
+    }
+    
+    class func loadImage(named name: String, fileExtension: String = "jpg") -> UIImage? {
```
nsobadjiev: If you look at how you ended up using this method, you'll see that it's mostly with force unwrapping like this:

```
GiniVisionTestsHelper.loadImage(named: "invoice")!
```

It might be better to return a non-optional here so you don't have to do that.
I agree that this is a better signature, but since it's a test helper, I think it's ok to do any of the following:
* Unwrap before returning
* Assert that the image exists and crash if it doesn't
* Return an empty image (not my favourite)
* Leave it as-is

kikettas: You are right, I think in this case is better to do the unwrapping inside of the method. It is also more consistent, since it is what I do in the rest of methods.

PR #284, opened 2018-09-12
Release 4.0.1

kikettas:
### Description
Release 4.0.1 version. It include the layout fixes for iOS 12, the extraction of the localizations into an enum and the test subspec.

### How to test
Nothing to test

### Merging
Automatic


Example/Example Swift/Info.plist
```diff
@@ -49,7 +49,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>0.9.0</string>
+	<string>0.9.1</string>
```
nsobadjiev: I think the example should soon reach version 1.

GiniVision/Classes/Core/Helpers/GiniVisionUtils.swift
```diff
@@ -31,7 +31,7 @@ func UIImageNamedPreferred(named name: String) -> UIImage? {
  
  - returns: String resource for the given key.
  */
-func NSLocalizedStringPreferred(_ key: String, comment: String, args: CVarArg? = nil) -> String {
+func NSLocalizedStringPreferredFormat(_ key: String, comment: String) -> String {
```
nsobadjiev: Why format? There's no format there.

kikettas: It is actually the format, because it could have external arguments, like this example: "This is an %@"

Example/Example Swift/Info.plist
kikettas: I think it is not a big deal since it is the example app, not the showcase app (no downloads so far)

PR #289, opened 2018-10-25
Documents metadata for networking plugin

kikettas:
### Description
Added document upload metadata support for the networking plugin. Additionally, it has been added to the Component API example as well.

### How to test
Run the example app, and check that when creating partial and composite documents, the metadata headers are added to the request (in both the Screen API and Component API)

### Merging
Manual

### DIFF
--- a/Example/Example Swift/ComponentAPIDocumentsService.swift
+++ b/Example/Example Swift/ComponentAPIDocumentsService.swift
@@ -15,9 +15,11 @@ final class ComponentAPIDocumentsService: ComponentAPIDocumentServiceProtocol {
     var partialDocuments: [String: PartialDocumentInfo] = [:]
     var compositeDocument: GINIDocument?
     var analysisCancellationToken: BFCancellationTokenSource?
+    var documentMetadata: GINIDocumentMetadata?
     
-    init(sdk: GiniSDK) {
+    init(sdk: GiniSDK, documentMetadata: GINIDocumentMetadata?) {
         self.giniSDK = sdk
+        self.documentMetadata = documentMetadata
     }
     
     func startAnalysis(completion: @escaping ComponentAPIAnalysisCompletion) {
@@ -130,6 +132,7 @@ extension ComponentAPIDocumentsService {
                 return self?.giniSDK.documentTaskManager.createPartialDocument(withFilename: fileName,
                                                                                from: document.data,
                                                                                docType: docType,
+                                                                               metadata: self?.documentMetadata,
                                                                                cancellationToken: cancellationToken)
             }).continueWith(block: { task in
                 if let createdDocument = task.result as? GINIDocument {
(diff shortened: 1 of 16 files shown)


kikettas:
@nsobadjiev There aren't breaking changes since the only `public` things that change are the 'GiniVision.viewController(...)' methods with the addition of the 'documentMetadata' (which has a default value). I thought about having a `GiniVisionNetworkingConfiguration`, but I think that it'd be an overkill.

nsobadjiev:
You don't have to have an explicit `GiniVisionNetworkingConfiguration` - you can use the standard configuration class.

kikettas:
@nsobadjiev I cannot, since the metadata would have to be in an extension of `GiniConfiguration` and stored properties are not allowed in extensions

Example/Example Swift/ComponentAPIDocumentsService.swift
```diff
@@ -15,9 +15,11 @@ final class ComponentAPIDocumentsService: ComponentAPIDocumentServiceProtocol {
     var partialDocuments: [String: PartialDocumentInfo] = [:]
     var compositeDocument: GINIDocument?
     var analysisCancellationToken: BFCancellationTokenSource?
+    var documentMetadata: GINIDocumentMetadata?
     
-    init(sdk: GiniSDK) {
+    init(sdk: GiniSDK, documentMetadata: GINIDocumentMetadata?) {
```
a-szotyori: Since document metadata is optional, I would add it as a property or create a second constructor. This would keep backwards compatibility, too.

Example/Example Swift/ScreenAPICoordinator.swift
```diff
@@ -25,25 +25,29 @@ final class ScreenAPICoordinator: NSObject, Coordinator {
...
     weak var analysisDelegate: AnalysisDelegate?
     var visionDocuments: [GiniVisionDocument]?
     var visionConfiguration: GiniConfiguration
     var sendFeedbackBlock: (([String: GINIExtraction]) -> Void)?
     
     init(configuration: GiniConfiguration,
          importedDocuments documents: [GiniVisionDocument]?,
-         client: GiniClient) {
+         client: GiniClient,
+         documentMetadata: GINIDocumentMetadata?) {
```
a-szotyori: Same as above.

GiniVision/Classes/Networking/DocumentsService.swift
```diff
@@ -14,8 +14,10 @@ final class DocumentsService: DocumentServiceProtocol {
     var partialDocuments: [String: PartialDocumentInfo] = [:]
     var compositeDocument: GINIDocument?
     var analysisCancellationToken: BFCancellationTokenSource?
+    var metadata: GINIDocumentMetadata?
     
-    init(sdk: GiniSDK) {
+    init(sdk: GiniSDK, metadata: GINIDocumentMetadata?) {
```
a-szotyori: Same as above.

GiniVision/Classes/Networking/Extensions/GiniScreenAPICoordinator+Networking.swift
```diff
@@ -72,7 +72,8 @@ extension GiniScreenAPICoordinator {
     
     convenience init(client: GiniClient,
                      resultsDelegate: GiniVisionResultsDelegate,
-                     giniConfiguration: GiniConfiguration) {
+                     giniConfiguration: GiniConfiguration,
+                     documentMetadata: GINIDocumentMetadata?) {
```
a-szotyori: Ditto.

PR #294, opened 2018-12-28
Improve multipage memory usage

kikettas:
### Description
After taking pictures from both the camera and pickers, they weren't downsized to their proper size in the multipage collection views, increasing the memory usage unnecessarily.
All the set-up logic for the cells has been moved to a new collection presenter, which take over this task now.

### How to test
Run the example app without in the `develop` branch, import 10 pictures and check the memory usage. Do the same in the current branch and compare the results

### Merging
Automatic


Example/GiniVision.xcworkspace/xcshareddata/WorkspaceSettings.xcsettings
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>BuildSystemType</key>
+	<string>Original</string>
```
kikettas: This is needed since there is a problem with Xcode 10 and Cocoapods which does not refresh the changes in the pod 😢.

GiniVision/Classes/Core/Screens/Multipage Review/MultipageReviewCollectionsCellPresenter.swift
```diff
@@ -0,0 +1,209 @@
...
+    fileprivate let giniConfiguration: GiniConfiguration
+    fileprivate let thumbnailsQueue = DispatchQueue(label: "Thumbnails queue")
+    
+    enum MultipageCollectionCellType {
+        case main(MultipageReviewMainCollectionCell, (NoticeActionType) -> Void)
+        case pages(MultipageReviewPagesCollectionCell)
+    }
+    
+    enum ThumbnailType {
+        case big, small
```
nsobadjiev: Here, since `big` is actually 100% scale (the full image), I'd call it original. Otherwise, I'd still expect it to be downscaled at least a little.
And since the big one is the original, it's not even a thumbnail, so I'd actually call them "original" and "thumbnail" and the enum "ImageScale" or something.

GiniVision/Classes/Core/Screens/Multipage Review/MultipageReviewCollectionsCellPresenter.swift
```diff
@@ -0,0 +1,209 @@
...
+        
+        var scale: CGFloat {
+            switch self {
+            case .big:
+                return 1.0
+            case .small:
+                return 1/4
```
nsobadjiev: My intuition is that 25% is still too much for a thumbnail. But maybe it gets distorted. Did you try lower values?

GiniVision/Classes/Core/Screens/Multipage Review/MultipageReviewCollectionsCellPresenter.swift
```diff
@@ -0,0 +1,209 @@
...
+    func setUp(cell: MultipageReviewPagesCollectionCell,
+               with page: GiniVisionPage,
+               fetchThumbnailTrigger: @autoclosure () -> Void,
+               pageIndex: Int,
+               selected: Bool) {
+        // Thumbnail set
+        if let thumbnail = self.thumbnails[page.document.id, default: [:]][.small] {
+            cell.documentImage.contentMode = thumbnail.size.width > thumbnail.size.height ?
+                .scaleAspectFit :
+                .scaleAspectFill
```
nsobadjiev: Doesn't `scaleAspectFit` work well for both kinds of ratio?

GiniVision/Classes/Core/Models/GiniVisionPage.swift
```diff
@@ -11,14 +11,23 @@ import Foundation
...
+    }
+    
     public init(document: GiniVisionDocument, error: Error? = nil, isUploaded: Bool = false) {
         self.document = document
         self.error = error
         self.isUploaded = isUploaded
     }
+    
+    public func isEqual(to element: GiniVisionPage) -> Bool {
+        return error?.localizedDescription == element.error?.localizedDescription &&
```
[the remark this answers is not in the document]
kikettas: Yeah, maybe the naming is a bit confusing. I can come up with a better name.

GiniVision/Classes/Core/Screens/Multipage Review/MultipageReviewCollectionsCellPresenter.swift
kikettas: It's not the original version in many cases (mostly on iPhone). It's the size that matched the screen width.

GiniVision/Classes/Core/Screens/Multipage Review/MultipageReviewCollectionsCellPresenter.swift
kikettas: Yes, I tried both 1/5 and 1/6 and it got distorted. Plus the performance with 1/4 is quite good.

GiniVision/Classes/Core/Screens/Multipage Review/MultipageReviewCollectionsCellPresenter.swift
kikettas: It doesn't look very well, mostly if the ratio width/height is too low. That's why is preferred to use AspectFill for portrait images.

GiniVision/Classes/Core/Models/GiniVisionPage.swift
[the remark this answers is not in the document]
nsobadjiev: Can you still avoid using descriptions as part of the comparison? Isbn't the error code enough?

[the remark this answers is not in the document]
kikettas: `Error` does not conform to `Equatable`, so I cannot compare the error itself

PR #295, opened 2018-12-31
Fixes image picker animation speed

kikettas:
### Description
When using old iOS devices, the image selection was really slow (there was a delay between the selection and the animation).

### How to test
Run the example app in an old device (iPad 3 Retina for instance), and check that the animation in the cells is fixed)

### Merging
Merge after #294


nsobadjiev:
CI fails @kikettas

PR #297, opened 2019-01-15
Add accounting api support

kikettas:
### Description
Added support for Accounting API.

### How to test
In the example app, just pass the `.accounting` type in the `GiniVision.viewController(...api:)` method.

### Merging
Automatic


Example/Podfile
```diff
@@ -5,7 +5,7 @@ platform :ios, '9.0'
 use_frameworks!
 
 def project_pods
-    pod 'Gini-iOS-SDK'
+    pod 'Gini-iOS-SDK', :git => 'git@github.com:gini/gini-sdk-ios.git' ,:tag => '1.2.0-beta.1'
```
nsobadjiev: Is the git URL actually needed?

GiniVision/Classes/Networking/AccountingDocumentService.swift
```diff
@@ -0,0 +1,139 @@
...
+        self.giniSDK.sessionManager.logIn()
+    }
+    
+    func cancelAnalysis() {
+        if let document = document {
+            delete(document)
+        }
+        
+        analysisCancellationToken?.cancel()
+        resetToInitialState()
```
nsobadjiev: Shouldn't these two lines be inside the if? Just for clarity?

Example/Podfile
kikettas: I thought it wasn't but didn't work for me locally

GiniVision/Classes/Networking/AccountingDocumentService.swift
kikettas: No, the analysis can be cancelled when the document is being upload.

Example/Podfile
nsobadjiev: Also, you should be using the HTTPS URL not the git one.

PR #306, opened 2019-01-29
Add tint color customization for Document Explorer screen

kikettas:
### Description
If using a custom UIAppearance in the host app, there were some issues that could affect the document explorer screen in the GVL for iOS >= 11.0. To fix this, a new property in the `GiniConfiguration` was added, allowing the tint color customization.

### How to test
Customize the UIAppearance for the UINavigationBar, UISearchBar and UIBarButtonItem, and launch the document explorer in the GVL to check that there is no issues.

### Merging
Automatic

### DIFF
--- a/GiniVision/Classes/Core/Screens/Document picker/DocumentPickerCoordinator.swift
+++ b/GiniVision/Classes/Core/Screens/Document picker/DocumentPickerCoordinator.swift
@@ -97,6 +97,9 @@ public final class DocumentPickerCoordinator: NSObject {
     
     let galleryCoordinator: GalleryCoordinator
     let giniConfiguration: GiniConfiguration
+    fileprivate lazy var navigationBarAppearance: UINavigationBar = .init()
+    fileprivate lazy var searchBarAppearance: UISearchBar = .init()
+    fileprivate lazy var barButtonItemAppearance: UIBarButtonItem = .init()
     
     fileprivate var acceptedDocumentTypes: [String] {
         switch giniConfiguration.fileImportSupportedTypes {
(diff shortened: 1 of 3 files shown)


GiniVision/Classes/Core/Screens/Document picker/DocumentPickerCoordinator.swift
```diff
@@ -172,6 +177,20 @@ public final class DocumentPickerCoordinator: NSObject {
         
         if #available(iOS 11.0, *) {
             documentPicker.allowsMultipleSelection = giniConfiguration.multipageEnabled
+            
+            // Starting on iOS 11.0, the UIDocumentPickerViewController navigation bar can't almost be customized,
+            // being only possible to customize the tint color. To avoid issues with custom UIAppearance styles,
+            // this is reset to default, saving the current state for further restoring when dismissing.
```
nsobadjiev: Starting **with** iOS 11.0, the UIDocumentPickerViewController navigation bar **almost** can't ~almost~ be customized, ~being~ only **being** possible to customize the tint color. To avoid issues with custom UIAppearance styles, this is reset to default, saving the current state ~for further restoring when dismissing~ in order to restore it during dismissal.

GiniVision/Classes/Core/Screens/Document picker/DocumentPickerCoordinator.swift
```diff
@@ -225,6 +244,64 @@ extension DocumentPickerCoordinator {
         
         return nil
     }
+    
+    @available(iOS 11.0, *)
+    func saveCurrentAppAppearance() {
```
nsobadjiev: It's not really the "App" appearance, is it? It's more like the navigation bar. I find the name misleading.

PR #313, opened 2019-02-11
Extract camera preview to a different view controller

kikettas:
### Description
All camera preview related functionality has been extracted from the `CameraViewController` to the new `CameraPreviewViewController`.

### How to test
Run the Example and check that everything was working as expected.

### Merging
Automatic

### DIFF
--- a/GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
+++ b/GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
@@ -163,7 +163,7 @@ final class CameraPreviewViewController: UIViewController {
         camera.captureStillImage(completion: { data, error in
             if let data = data,
                 let image = UIImage(data: data),
-                let imageData = image.jpegData(compressionQuality: 0.9) {
+                let imageData = image.jpegData(compressionQuality: 1.0) {
                 completion(imageData, error)
             } else {
                 completion(data, error)
--- a/GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
+++ b/GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
@@ -1,0 +1,368 @@
+//
+//  CameraPreviewViewController.swift
+//  GiniVision
+//
+//  Created by Enrique del Pozo Gómez on 8/10/18.
+//
+
+import UIKit
+import AVFoundation
+
+protocol CameraPreviewViewControllerDelegate: class {
+    func cameraPreview(viewController: CameraPreviewViewController,
+                       didDetect qrCodeDocument: GiniQRCodeDocument)
+}
+
+final class CameraPreviewViewController: UIViewController {
+    
+    weak var delegate: CameraPreviewViewControllerDelegate?
+    
+    var documentDetector: DocumentDetectorProtocol?
+    fileprivate let giniConfiguration: GiniConfiguration
+    fileprivate typealias FocusIndicator = UIImageView
+    fileprivate var camera: Camera?
+    fileprivate var cameraState = CameraState.notValid
+    fileprivate var documentDetectedInfo: DetectedDocumentInfo?
+    fileprivate var defaultImageView: UIImageView?
+    fileprivate var focusIndicatorImageView: UIImageView?
+    fileprivate let interfaceOrientationsMapping: [UIInterfaceOrientation: AVCaptureVideoOrientation] = [
+        .portrait: .portrait,
+        .landscapeRight: .landscapeRight,
+        .landscapeLeft: .landscapeLeft,
+        .portraitUpsideDown: .portraitUpsideDown
+    ]
+    
+    fileprivate var cameraFocusSmall: UIImage? {
+        return UIImageNamedPreferred(named: "cameraFocusSmall")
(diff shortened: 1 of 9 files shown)


GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
```diff
@@ -0,0 +1,292 @@
...
+        
+        setupCamera(giniConfiguration: giniConfiguration)
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    deinit {
+        NotificationCenter.default.removeObserver(self)
```
nsobadjiev: Is this really necessary? Wouldn't the notification center be smart enough to deal with it alone?

GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
```diff
@@ -0,0 +1,292 @@
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        updatePreviewViewOrientation() // Video orientation should be updated once the view has been loaded
+    }
+    
+    public override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        camera?.start()
```
nsobadjiev: Can you make camera non-optional? Force-unwrapped variable might also be fine. If the camera turns out to be nil, the app better crash. Just IMHO.

GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
```diff
@@ -0,0 +1,292 @@
...
+    public override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        camera?.stop()
+    }
+    
+    public override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+        super.viewWillTransition(to: size, with: coordinator)
+        
+        coordinator.animate(alongsideTransition: { [weak self] _ in
+            guard let `self` = self else {
```
nsobadjiev: Is this guard necessary? Why can't you just have `self?.updatePreviewViewOrientation()`?

GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
```diff
@@ -0,0 +1,292 @@
...
+                        #endif
+                    }
+                }
+            }
+        }
+                
+        if giniConfiguration.qrCodeScanningEnabled {
+            camera?.setupQRScanningOutput()
+            camera?.didDetectQR = { [weak self] qrDocument in
+                guard let `self` = self else { return }
```
nsobadjiev: Same here.

GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
```diff
@@ -0,0 +1,292 @@
...
+                let imageData = image.jpegData(compressionQuality: 0.2) {
+                completion(imageData, nil)
+            }
+            #endif
+        }
+        
+        camera.captureStillImage(completion: { data, error in
+            if let data = data,
+                let image = UIImage(data: data),
+                let imageData = image.jpegData(compressionQuality: 0.9) {
```
nsobadjiev: Why 0.9? Wasn't the optimal compression level 0.7? Or maybe something changed.

GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
kikettas: Without this it doesn't look very well, since the animation doesn't happen at the same time

GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
kikettas: Definitely makes sense, I just copied and paster most of the stuff from one file to the other and didn't consider making many changes tbh.

GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
kikettas: Yes, unfortunately it has to be disposed manually.

GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
kikettas: it was always 0.9, but I would make it 1.0 since it doesn't affect the image size that much

GiniVision/Classes/Core/Screens/Camera/CameraPreviewViewController.swift
kikettas: Not here, since the delegate needs it in the line below

nsobadjiev: Why can't you write `self?.delegate?.cameraPreview(self, didDetect: qrDocument)`

PR #314, opened 2019-02-12
Make the supported formats screen optional

kikettas:
### Description
Added flag in the `GiniConfiguration` to disable the Supported formats screen.

### How to test
Set the `shouldShowSupportedFormatsScreen` to false and check that the screen is not shown.

### Merging
Automatic

### DIFF
--- a/GiniVision/Classes/Core/Screens/Help/HelpMenuViewController.swift
+++ b/GiniVision/Classes/Core/Screens/Help/HelpMenuViewController.swift
@@ -18,14 +18,53 @@ final public class HelpMenuViewController: UITableViewController {
     let giniConfiguration: GiniConfiguration
     let tableRowHeight: CGFloat = 64
     var helpMenuCellIdentifier = "helpMenuCellIdentifier"
-    lazy var items: [(text: String, id: Int)] = {
-        var items: [(text: String, id: Int)] = [
-            (.localized(resource: HelpStrings.menuFirstItemText), 1),
-            (.localized(resource: HelpStrings.menuThirdItemText), 3)
-        ]
+    
+    enum Item {
+        case noResultsTips
+        case openWithTutorial
+        case supportedFormats
         
-        if self.giniConfiguration.openWithEnabled {
-            items.insert((.localized(resource: HelpStrings.menuSecondItemText), 2), at: 1)
+        var title: String {
+            switch self {
+            case .noResultsTips:
+                return .localized(resource: HelpStrings.menuFirstItemText)
+            case .openWithTutorial:
+                return .localized(resource: HelpStrings.menuSecondItemText)
+            case .supportedFormats:
+                return .localized(resource: HelpStrings.menuThirdItemText)
+            }
+        }
+        
+        var viewController: UIViewController {
+            let viewController: UIViewController
+            switch self {
+            case .noResultsTips:
+                let title: String = .localized(resource: ImageAnalysisNoResultsStrings.titleText)
+                let topViewText: String = .localized(resource: ImageAnalysisNoResultsStrings.warningHelpMenuText)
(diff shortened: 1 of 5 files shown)


GiniVision/Classes/Core/Screens/Help/HelpMenuViewController.swift
```diff
@@ -135,9 +174,24 @@ extension HelpMenuViewController {
...
-            navigationController?.pushViewController(viewController, animated: true)
+        let item = items[indexPath.row]
+        let viewController = item.viewController
+        viewController.setupNavigationItem(usingResources: backToMenuButtonResource,
+                                           selector: #selector(back),
+                                           position: .left,
+                                           target: self)
+        
+        if let imageNoResultsViewController = viewController as? ImageAnalysisNoResultsViewController {
+            imageNoResultsViewController.didTapBottomButton = {
```
nsobadjiev: Here, you ruin the "magic" of encapsulating the type of view controller. Can you find a way to setup the `ImageAnalysisNoResultsViewController` in a way that it won't break the abstraction?

GiniVision/Classes/Core/Screens/Help/HelpMenuViewController.swift
```diff
@@ -135,9 +174,24 @@ extension HelpMenuViewController {
...
+                if let cameraViewController = (self.navigationController?
+                    .viewControllers
+                    .compactMap { $0 as? CameraViewController })?
```
nsobadjiev: You know how much I hate iterating view controller in order to find a specific one. It seems like a task for a Coordinator. View controllers shouldn't know about the hierarchy.

GiniVision/Classes/Core/Screens/Help/HelpMenuViewController.swift
```diff
@@ -135,9 +174,24 @@ extension HelpMenuViewController {
...
+                    .first {
+                    _ = self.navigationController?.popToViewController(cameraViewController, animated: true)
+                }
+            }
         }
+        navigationController?.pushViewController(viewController, animated: true)
```
nsobadjiev: You also assume a navigation controller setup. Again, something a coordinator should do.

GiniVision/Classes/Core/Screens/Help/HelpMenuViewController.swift
kikettas: You are right, and I think we already discuss this in the past. The reason why we didn't use a coordinator here is that the `HelpMenuViewController` is a public screen and that would mean exposing this coordinator.

One thing that we could do to not introduce breaking changes is to add a delegate for those actions that require a coordinator (like navigation), and fallback to the current implementation.

What do you think?

nsobadjiev: That would work for me.

GiniVision/Classes/Core/Screens/Help/HelpMenuViewController.swift
kikettas: Can't make it with the new `HelpMenuViewControllerDelegate`, but since it's marked as deprecated internally, it shouldn't be a problem anymore.

GiniVision/Classes/Core/Screens/Help/HelpMenuViewController.swift
kikettas: Fixed in the `HelpMenuViewControllerDelegate`

GiniVision/Classes/Core/Screens/Help/HelpMenuViewController.swift
nsobadjiev: When? it's going to take a long time until we actually remove it.

PR #316, opened 2019-02-19
Feature flash mode

kikettas:
### Description
Added flash toggle in the camera screen. It'll be positioned differently depending on the device, being like this:

image
 
image

### How to test
Run the example app and check that all the camera buttons are shown. Change the `GiniConfiguration` properties to see that the buttons position change.

### Merging
Automatic


CameraButtonsViewControllerTests.swift
```diff
@@ -0,0 +1,282 @@
...
+    let delegateMock = CameraButtonViewControllerDelegateMock()
+    lazy var imageData: Data = {
+        let image = GiniVisionTestsHelper.loadImage(named: "invoice")
+        let imageData = image.jpegData(compressionQuality: 0.9)!
+        return imageData
+    }()
+    
+    override func setUp() {
+        super.setUp()
+        giniConfiguration = GiniConfiguration.shared
```
nsobadjiev: Why are you using the global configuration in tests? Why not create a custom one?

CameraButtonsViewControllerTests.swift
```diff
@@ -0,0 +1,282 @@
...
+        giniConfiguration.multipageEnabled = true
+        cameraButtonsViewController = CameraButtonsViewController(giniConfiguration: giniConfiguration,
+                                                                  isFlashSupported: true,
+                                                                  currentDevice: IpadDevice())
+        _ = cameraButtonsViewController.view
+        cameraButtonsViewController.addFileImportButton()
+        
+        let innerRightVerticalStackview = cameraButtonsViewController.rightStackView
+            .arrangedSubviews.first as? UIStackView
+        let innerLeftVerticalStackview = cameraButtonsViewController.leftStackView
```
nsobadjiev: This and the previous two methods duplicate these lines of code. Maybe you can extract it as a method?

GiniVision/Classes/Core/Screens/Camera/CameraViewController.swift
```diff
@@ -179,8 +134,9 @@ public typealias CameraScreenFailureBlock = (_ error: GiniVisionError) -> Void
      
      - returns: A view controller instance allowing the user to take a picture or pick a document.
      */
-    public init(giniConfiguration: GiniConfiguration) {
+    public init(giniConfiguration: GiniConfiguration, currentDevice: UIDevice = .current) {
```
nsobadjiev: Is there a particular reason you chose to inject the device via the constructor? I'm not sure it would be better, but maybe just a settable property would be enough. As far as I understand it's only there because of testing.

PR #322, opened 2019-03-29
Add doc type to initializers and methods.

kikettas:
### Description
Add the option to specify the doc type hint in the GVL initializers.

### How to test
Specify the docType in the GVL initializers and check that it's propagated to the Gini API SDK

### Merging
Automatic


GiniVision/Classes/Networking/AccountingDocumentService.swift
```diff
@@ -6,19 +6,21 @@
 //
 
 import Foundation
-import Gini_iOS_SDK
+import Gini_iOS_SDK
```
a-szotyori: Why is this import needed now?

kikettas: It was already there. Not sure why the repo changed this file at that lien, myabe a line break or similar.

PR #326, opened 2019-04-10
Integrate the new API SDK

kikettas:
### Description
Integration of the new API SDK. Additionally, the deployment target has been bumped to iOS 10.0 and the Obj-C support has been removed in the networking plugin.

### How to test
Run the example app and test that everything works as previously.

### Merging
Automatic


Example/Example Swift/ComponentAPICoordinator.swift
```diff
@@ -335,10 +335,9 @@ extension ComponentAPICoordinator {
                 case .success(let extractions):
                     self.handleAnalysis(with: extractions)
                 case .failure(let error):
-                    let error = error as? AnalysisError ?? AnalysisError.unknown
-                    guard error != .cancelled else { return }
+                    guard error != .requestCancelled else { return }
 
-                    self.showErrorInAnalysisScreen(with: error.message) {
+                    self.showErrorInAnalysisScreen(with: AnalysisError.unknown.message) {
```
nsobadjiev: So you disregard the `error` object and always show an "unknown" error? Are you sure that's fine? Can you at least log it?

Example/Example Swift/ComponentAPICoordinator.swift
```diff
@@ -682,9 +681,9 @@ extension ComponentAPICoordinator {
 
 extension ComponentAPICoordinator {
     
-    fileprivate func handleAnalysis(with extractions: [String: GINIExtraction]) {
+    fileprivate func handleAnalysis(with extractions: [Extraction]) {
         let payFive = ["paymentReference", "iban", "bic", "paymentReference", "amountToPay"]
-        let hasPayFive = extractions.filter { payFive.contains($0.0) }.count > 0
+        let hasPayFive = extractions.filter { payFive.contains($0.name ?? "no-name") }.count > 0
```
nsobadjiev: So basically, if name is null, you compare against "no-name" which is guaranteed to fail? It's a little confusing. I'd rather write

`$0.name ? payFive.contains($0.name) : false`

It's still not super concise but it's easier to follow as logic.

Example/Example Swift/ComponentAPIDocumentsService.swift
```diff
@@ -57,36 +55,24 @@ final class ComponentAPIDocumentsService: ComponentAPIDocumentServiceProtocol {
...
+    func sendFeedback(with updatedExtractions: [Extraction]) {
+        guard let document = document else { return }
+        documentService.submitFeedback(for: document, with: updatedExtractions) { result in
+            switch result {
+            case .success:
                 print("🚀 Feedback sent with \(updatedExtractions.count) extractions")
-                return nil
-            })
+            case .failure(let error):
+                let message = "❌ Error sending feedback for document with id: \(document.id) error: \(error)"
```
nsobadjiev: I don't see a benefit in having a local var for that.

Example/Example Swift/ComponentAPIDocumentsService.swift
```diff
@@ -121,148 +109,93 @@ final class ComponentAPIDocumentsService: ComponentAPIDocumentServiceProtocol {
...
+                                       docType: docType,
+                                       type: .partial(document.data),
+                                       metadata: metadata) { result in
+                                        switch result {
+                                        case .success(let createdDocument):
+                                            print("📄 Created document with id: \(createdDocument.id) " +
+                                                "for vision document \(document.id)")
+                                            completion(.success(createdDocument))
+                                        case .failure(let error):
+                                            print("❌ Document creation failed")
```
nsobadjiev: Why not also print the error?

Example/Example Swift/ComponentAPIDocumentsService.swift
```diff
@@ -121,148 +109,93 @@ final class ComponentAPIDocumentsService: ComponentAPIDocumentServiceProtocol {
...
-                if task.isCancelled || task.error != nil {
-                    print("❌ Error deleting partial document with id: \(id)")
-                } else {
-                    print("🗑 Deleted partial document with id: \(id)")
-                }
-                
-                return nil
-            })
-        
+    func delete(partial document: Document) {
```
nsobadjiev: The two delete methods are nearly identical. I'd merge them (even if it means that the logs don't specify if it was a partial or composite)

Example/Example Swift/ComponentAPIDocumentsService.swift
```diff
@@ -121,148 +109,93 @@ final class ComponentAPIDocumentsService: ComponentAPIDocumentServiceProtocol {
...
+                                case .success(let createdDocument):
+                                    print("🔎 Starting analysis for composite document with id \(createdDocument.id)")
+                                    
+                                    self.analysisCancellationToken = CancellationToken()
+                                    self.documentService
+                                        .extractions(for: createdDocument,
+                                                     cancellationToken: self.analysisCancellationToken!,
+                                                     completion: self.handleResults(completion: completion))
+                                case .failure(let error):
+                                    print("❌ Composite document creation failed")
```
nsobadjiev: Again, it might be interesting to print the error.

Example/Example Swift/ComponentAPIDocumentsService.swift
```diff
@@ -121,148 +109,93 @@ final class ComponentAPIDocumentsService: ComponentAPIDocumentServiceProtocol {
...
+                completion(.success(extractions))
+            case .failure(let error):
+                switch error {
+                case .requestCancelled:
                     print("❌ Cancelled analysis process")
-                    completion(.failure(AnalysisError.cancelled))
-                    
-                    return BFTask<AnyObject>.cancelled()
+                default:
+                    print("❌ Finished analysis process with error: \(error)")
```
nsobadjiev: Like you print it here already ;)

Example/Example Swift/CredentialsManager.swift
```diff
@@ -24,12 +24,12 @@ final class CredentialsManager {
...
-                              clientSecret: client_password,
-                              clientEmailDomain: client_email_domain)
+            return Client(id: client_id,
+                          secret: client_password,
+                          domain: client_email_domain)
         }
-        return GiniClient(clientId: "",
-                          clientSecret: "",
-                          clientEmailDomain: "")
+        return Client(id: "",
```
nsobadjiev: I think I did the same in the past and Peter complained that I'm returning an empty Client and not a null or something. I tend to agree. A Client object with empty strings is not super useful.

Example/Example Swift/ResultTableViewController.swift
```diff
@@ -16,12 +16,12 @@ import Gini_iOS_SDK
...
     
     /**
-     The result dictionary from the analysis process.
+     The result collection from the analysis process.
      */
-    var result: [String: GINIExtraction]!
-    
-    fileprivate var sortedKeys: [String] {
-        return Array(result.keys).sorted(by: <)
+    var result: [Extraction]! {
```
nsobadjiev: Hmmm... why the implicitly unwrapped optional?

GiniVision/Classes/Networking/AccountingDocumentService.swift
```diff
@@ -77,70 +87,52 @@ final class AccountingDocumentService: DocumentServiceProtocol {
...
+                                       metadata: metadata) { result in
+                                        switch result {
+                                        case .success(let createdDocument):
+                                            Log(message: "Created document with id: \(createdDocument.id) " +
+                                                "for vision document \(document.id)", event: "📄")
+                                            
+                                            self.document = createdDocument
+                                            completion(.success(createdDocument))
+                                        case .failure(let error):
+                                            Log(message: "Document creation failed", event: .error)
```
nsobadjiev: Here, you use a `Log` and other places you use `print`. I assume everything should be `Log`?

Example/Example Swift/ComponentAPICoordinator.swift
kikettas: Before doing this I checked the trace of the error and it was always showing this error, so didn't want to make it more complicated. 
It's already logged in the networking side

Example/Example Swift/ComponentAPIDocumentsService.swift
kikettas: It was like that because it was a longer line before. Now it makes sense to print directly 👍

Example/Example Swift/ComponentAPIDocumentsService.swift
kikettas: Makes sense. I was only doing it for the logs, but I can print the `sourceClassification` instead

Example/Example Swift/CredentialsManager.swift
kikettas: We actually discussed this already. It was an optional before and we changed it to this so the example app still works without credentials (avoiding bad first impressions).
Adding a warning there would help though.

Example/Example Swift/ResultTableViewController.swift
kikettas: 🤷‍♂️ No idea to be honest, maybe I _copied-pasted_ and didn't notice it

GiniVision/Classes/Networking/AccountingDocumentService.swift
kikettas: No. I only use _Log_ within the library. The _print_s are part of the Example app.

Example/Example Swift/ComponentAPIDocumentsService.swift
nsobadjiev: You missed this one.

PR #327, opened 2019-04-12
Add credentials removal method to GiniVision class.

kikettas:
### Description
Add credentials removal method to `GiniVision` class.

### How to test
Call the recently created method and check that the stored credentials are removed

### Merging
Automatic


nsobadjiev:
Build fails ;(

PR #328, opened 2019-04-12
Add EPS QR code support

kikettas:
### Description
Add EPS QR code support (https://members.stuzza.at/eservice/shops/eps4mobile/demo/index.php)

### How to test
Generate a QR code in the website linked above, and scan it.

### Merging
Merge after #327


GiniVision/Classes/Networking/Extensions/GiniScreenAPICoordinator+Networking.swift
```diff
@@ -203,6 +203,20 @@ extension GiniScreenAPICoordinator: GiniVisionDelegate {
     }
     
     func didCapture(document: GiniVisionDocument, networkDelegate: GiniVisionNetworkDelegate) {
+        // The EPS QR codes are a special case, since they cannot be analyzed by the Gini API and therefore, they are
```
nsobadjiev: I wouldn't say they "cannot" but "they don't have to be"

kikettas: Better, yeah

PR #335, opened 2019-05-06
Add flash ON by default flag

kikettas:
### Description
Add the flash ON by default flag in the `GiniConfiguration`.

### How to test
Disable the flash on by default in the `AppCoordinator` and check the flash is disabled on start.

### Merging
Automatic

### DIFF
--- a/GiniVision/Classes/Core/GiniConfiguration.swift
+++ b/GiniVision/Classes/Core/GiniConfiguration.swift
@@ -336,7 +336,13 @@ import UIKit
      
      */
     @objc public var flashToggleEnabled = false
-
+    
+    /**
+     When the flash toggle is enabled, this flag indicates whether
+     if the flash is on by default.
+     */
+    @objc public var flashOnByDefault = true
+    
     /**
      Sets the title text in the navigation bar on the camera screen.
      - Warning: Deprecated, use the ginivision.navigationbar.camera.title localized string instead
--- a/GiniVision/Classes/Core/GiniConfiguration.swift
+++ b/GiniVision/Classes/Core/GiniConfiguration.swift
@@ -338,8 +338,7 @@ import UIKit
     @objc public var flashToggleEnabled = false
     
     /**
-     When the flash toggle is enabled, this flag indicates whether
-     if the flash is on by default.
+     When the flash toggle is enabled, this flag indicates if the flash is on by default.
      */
     @objc public var flashOnByDefault = true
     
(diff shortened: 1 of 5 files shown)


GiniVision/Classes/Core/GiniConfiguration.swift
```diff
@@ -336,7 +336,13 @@ import UIKit
      
      */
     @objc public var flashToggleEnabled = false
-
+    
+    /**
+     When the flash toggle is enabled, this flag indicates whether
+     if the flash is on by default.
```
nsobadjiev: either "whether or not" or only "if"

GiniVision/Classes/Core/Screens/Camera/Camera.swift
```diff
@@ -44,7 +44,7 @@ final class Camera: NSObject, CameraProtocol {
         return videoDeviceInput?.device.hasFlash ?? false
         #endif
     }()
-    var isFlashOn: Bool = true
+    lazy var isFlashOn: Bool = giniConfiguration.flashOnByDefault
```
nsobadjiev: I'm not sure you really benefit from the laziness. If you just put it there to delay the setting, it's maybe not the best solution.

GiniVision/Tests/CameraButtonsViewControllerTests.swift
```diff
@@ -8,6 +8,7 @@
 import XCTest
 @testable import GiniVision
 
+// swiftlint:disable type_body_length
```
nsobadjiev: You can consider splitting the tests rather than having a linter exception (if you want).

GiniVision/Tests/CameraButtonsViewControllerTests.swift
```diff
@@ -279,4 +280,29 @@ final class CameraButtonsViewControllerTests: XCTestCase {
...
+    func testFlashStatusWhenOffByDefault() {
+        let giniConfiguration = GiniConfiguration()
+        giniConfiguration.flashOnByDefault = false
+        
+        cameraButtonsViewController = CameraButtonsViewController(giniConfiguration: giniConfiguration,
+                                                                  isFlashSupported: true,
+                                                                  currentDevice: IphoneDevice())
+        
+        XCTAssertFalse(cameraButtonsViewController.flashToggleButton.isSelected,
+                       "the flash toggle should be selected when flash in on by default")
```
nsobadjiev: "Off" by default. And since it's an error caused by copy-pasting, maybe you should generalize the whole method.

GiniVision/Tests/CameraButtonsViewControllerTests.swift
kikettas: I think it's fine like this

GiniVision/Tests/CameraButtonsViewControllerTests.swift
kikettas: For only two test it's an overkill

PR #379, opened 2019-12-17
PIA-391: Implement the return assistant screen on iOS

maciejtrybilo:
### Description

The return assistant screen. Missing localisation support and proper dark mode. Could think about some more tests.

### How to test

It's displayed instead of the analysis screen in the Screen API flow with some mock data. 

### Merging
Automatic


cursork:
I'm happy, and I've got it running locally.

PR #380, opened 2020-01-09
PIA-410: Line item details screen for the return assistant.

maciejtrybilo:


GiniVision/Classes/Core/Screens/Digital Invoice/GiniTextField.swift
```diff
@@ -0,0 +1,219 @@
...
+    }
+    
+    var textFieldLastBaselineAnchor: NSLayoutYAxisAnchor {
+        return textField.lastBaselineAnchor
+    }
+    
+    var keyboardType: UIKeyboardType {
+        
+        set {
+            textField.keyboardType = newValue
```
cursork: I'm kind of indecisive about all these fields that essentially just delegate to a private member like `textField` here... At some point, there's not much difference between just having it public and this approach? But I've also nothing against it - just a thought rather than anything to fix.

maciejtrybilo: I thought about it as well. It might change as the type evolves, but at this point I have a slight preference towards tightly controlling the API. For instance `textFont` and `textColor` tie the `textField` and the `prefixLabel` together  and I think it simplifies things for the client code.

cursork:
looks reasonable enough to me

PR #385, opened 2020-01-29
PIA-457: Fix camera flash not appearing on the first run.

maciejtrybilo:


GiniVision/Classes/Core/Screens/Camera/Camera.swift
```diff
@@ -41,7 +41,7 @@ final class Camera: NSObject, CameraProtocol {
         #if targetEnvironment(simulator)
         return true
         #else
-        return videoDeviceInput?.device.hasFlash ?? false
+        return videoDeviceInput?.device.hasFlash ?? true
```
a-szotyori: This might cause false positives on iPads without flash. Is there a way to postpone this check until `videoDeviceInput` is available? Or it might be enough to check which iPad versions we support and if all of them have a flash, then it's alright like this.

maciejtrybilo: You're right, better fix now.

PR #388, opened 2020-02-06
PIA-469: Fix pod lint for GVL 5

maciejtrybilo:
Explanation in ticket https://ginis.atlassian.net/browse/PIA-469

### DIFF
--- a/GiniVision.podspec
+++ b/GiniVision.podspec
@@ -18,6 +18,7 @@ The Gini Vision Library for iOS provides functionality to capture documents with
   s.social_media_url = 'https://twitter.com/gini'
   s.swift_version    = '5.0'
   s.ios.deployment_target = '10.0'
+  s.xcconfig = { 'ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES' => 'YES' }
   s.default_subspec = 'Core'
 
   s.subspec 'Core' do |core|
--- a/GiniVision.podspec
+++ b/GiniVision.podspec
@@ -18,7 +18,11 @@ The Gini Vision Library for iOS provides functionality to capture documents with
   s.social_media_url = 'https://twitter.com/gini'
   s.swift_version    = '5.0'
   s.ios.deployment_target = '10.0'
+  
+  # workaround for https://github.com/CocoaPods/CocoaPods/issues/9521
+  # Remove after iOS 11 is dropped.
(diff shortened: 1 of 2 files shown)


GiniVision/Tests/HelpMenuViewControllerTests.swift
```diff
@@ -93,10 +93,13 @@ final class HelpMenuViewControllerTests: XCTestCase {
         
         XCTAssertEqual(itemText, cell.textLabel?.text,
                        "cell text in the first row should be the same as the first item text")
+        
         // Compare cgColors because since iOS 13 cell.backgroundColor has the UIDynamicSystemColor type
-        XCTAssertEqual(cellBackgroundColor.cgColor,
-                       cell.backgroundColor!.cgColor,
-                       "cell background color should always be white")
+        if #available(iOS 13, *) {
```
a-szotyori: Shouldn't there be an `else` branch to assert this also on older versions?

GiniVision.podspec
```diff
@@ -18,6 +18,11 @@ The Gini Vision Library for iOS provides functionality to capture documents with
   s.social_media_url = 'https://twitter.com/gini'
   s.swift_version    = '5.0'
   s.ios.deployment_target = '10.0'
+  
+  # workaround for https://github.com/CocoaPods/CocoaPods/issues/9521
+  # Remove after iOS 11 is dropped.
+  s.xcconfig = { 'ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES' => 'YES' }
```
a-szotyori: What effect will this have on client app's ipa size? 🤔

GiniVision/Tests/HelpMenuViewControllerTests.swift
maciejtrybilo: This is a little flaky as for instance on iOS 10 the colour is in a different colour space and the comparison fails while on iOS 11 `cell.backgroundColor` is nil and the test crashes. I have another PR in preparation that changes the test sim to iPhone 11, so the test will run.

GiniVision.podspec
maciejtrybilo: This is what normally should happen. This line just forces it. You can see the setting in Build Settings. I'm still looking for ways to fix it without this line, but it works without side effects as far as I can tell.

PR #424, opened 2020-10-06
Pia 860

zladzeyka:
### Description
https://ginis.atlassian.net/browse/PIA-860

### Merging
Automatic

### DIFF
--- a/GiniVision/Classes/Networking/AnalysisResult.swift
+++ b/GiniVision/Classes/Networking/AnalysisResult.swift
@@ -19,8 +19,11 @@ import Gini
      */
     public let extractions: [String: Extraction]
     
-    public init(extractions: [String: Extraction], images: [UIImage]) {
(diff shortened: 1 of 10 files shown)


GiniVision/Classes/Networking/AnalysisResult.swift
```diff
@@ -19,8 +19,11 @@ import Gini
      */
     public let extractions: [String: Extraction]
```
a-szotyori, line 21:
```suggestion
    
    /*
    *  Extraction candidates dictionary. To get the candidates for an extraction look for the
    *  `Extraction.candidates` name in the dictionary. For example the IBAN extraction's `candidates` field 
    *  contains `"ibans"` and if you search for that in this dictionary, then you'll get all the IBAN candidates.
    */
```

a-szotyori (approved):

PR #427, opened 2020-10-12
Pia 855

zladzeyka:
### Description
Customizations for light/dark modes

### Merging
Automatic

### Issues
https://ginis.atlassian.net/browse/PIA-855


GiniVision/Classes/Core/Extensions/UIColor.swift
```diff
@@ -0,0 +1,27 @@
...
+//  UIColor.swift
+//  GiniVision
+//
+//  Created by Nadya Karaban on 12.10.20.
+//
+
+import Foundation
+
+extension UIColor {
+    func colorFromGiniColor(giniColor: GiniColor) -> UIColor {
```
a-szotyori, line 11: We could make this static and also rename it to better adhere to the swift method naming style:
```suggestion
    static func from(giniColor: GiniColor) -> UIColor {
```
By making it static we don't have to create a throwaway UIColor instance just to call this method.

GiniVision/Classes/Core/Screens/Document picker/Gallery/ImagePickerViewController.swift
```diff
@@ -67,11 +67,7 @@ final class ImagePickerViewController: UIViewController {
         
         title = currentAlbum.title
         
-        if #available(iOS 13.0, *) {
-            collectionView.backgroundColor = .systemBackground
-        } else {
-            view.backgroundColor = .white
-        }
+        view.backgroundColor = UIColor().colorFromGiniColor(giniColor: giniConfiguration.galeryScreenBackgroundColor)
```
a-szotyori, line 70: By renaming and making the method static we can call it like this:
```suggestion
        view.backgroundColor = UIColor.from(giniColor: giniConfiguration.galeryScreenBackgroundColor)
```

GiniVision/Classes/Core/GiniConfiguration.swift
```diff
@@ -8,6 +8,16 @@
 
 import UIKit
 
+@objc public class GiniColor : NSObject {
+    var lightModeColor: UIColor = .white
```
a-szotyori, line 12: Since we don't have an initializer with no arguments we don't have initialize the variables. They will be always get a value in the initializer.

a-szotyori (changes requested):
Nicely done and I found only a couple of things to adjust.

GiniVision/Classes/Core/Extensions/UIColor.swift
a-szotyori: What do you think of renaming it to `from(giniColor:)`? Usually having only the parameter name is enough, like CGRect's [`frame(withWidth: ...)`](https://developer.apple.com/documentation/coregraphics/cgrect/2903485-frame)

PR #460, opened 2021-08-30
Bugfix/pia 1477 improve file picker

a-szotyori:
### Description
If none of the picked files can be converted to `GiniVisionDocuments` then a new `DocumentPickerCoordinatorDelegate` method is called. In this case an error dialog is shown with the new error `FilePickerError.failedToOpenDocument`.

### How to test
Import the pdfs from the ticket and verify that the `good-...pdf` works and the `bad-...pdf` results in an alert dialog saying that the document couldn't be opened.

### Merging
Automatic

### Issues
https://ginis.atlassian.net/browse/PIA-1477


zladzeyka (approved):
Looks great and works great!

zladzeyka (approved):

